Issue #560, opened 2021-02-25
Remove tracing package


darioAnongba:
Remove the tracing package from the codebase.

PR #724, opened 2021-11-01
560 remove tracing

amkCha:
## Fixed Issue(s)
#560 remove tracing package

## Changelog

I'm not sure if I should add a comment for v21.10.0 in one of the sections : Features, breaking changes, bug fixes
So lmk :)


julien-marchand:
Close #560

PR #745, opened 2021-11-22
774 nested tenants

ggarri:
## PR description
- Orchestrate custom claims has been restored and refactored
- Support for nested tenants using `:` as separator char. For instance a user with access to `tenantOne:subTenantOne` will have access to all resources of `tenantOne` and resources of `tenantOne:subTenantOne`
- Include `username` as part of Custom Claims and additionally on `sub` using `:` as separator. For instance by using `tenantOne:username`
- Support for `ownerID` over resources such as Chains, Accounts and Transaction Schedules
- Resources owned by a specific `username`(ownerID) are only accessible by that user
- Refactor over multitenancy feature to reduce duplicity and aggregate request user info into an encapsulated object `UserInfo`
- Support for username impersonates ONLY on API-KEY
- 
## Fixed Issue(s)
#774

## Changelog


pkg/toolkit/app/http/middleware/metrics/metrics.go
```diff
@@ -54,8 +53,19 @@ func (m *Metrics) Handler(h http.Handler) http.Handler {
 func (m *Metrics) ServeHTTP(rw http.ResponseWriter, req *http.Request, next http.Handler) {
 	authLabels := append(
 		m.baseLabels,
-		"tenant_id", multitenancy.TenantIDFromContext(req.Context()),
+		"tenant_id", "",
 	)
+	
+	// if userInfo != nil {
```
6l20, line 59: We should remove commented code

services/api/store/postgres/migrations/19_replate_tenant_by_owner.go
```diff
@@ -0,0 +1,63 @@
+package migrations
```
6l20, line 1: File name should be 19_replace... ?

services/tx-sender/service/listener_test.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	"github.com/Shopify/sarama"
 	"github.com/cenkalti/backoff/v4"
+	envelope2 "github.com/consensys/orchestrate/pkg/utils/envelope"
```
6l20, line 12: Why envelope2 ?

6l20 (approved):

pkg/toolkit/app/http/middleware/metrics/metrics.go
ggarri: indeed, that was forgotten. Thanks

services/tx-sender/service/listener_test.go
ggarri: This is the automatic behaviour of GOlang adding dependencies with collisions

Issue #743, opened 2021-11-15
Tx-sender crashes when RPC node is down


ggarri:
### Describe the bug

tx-sender exits due to fail to update job to FAILED when rpc node is down

### Steps to reproduce

- Send transaction
- RPC node is down

**Actual result**

tx-sender exists

**Expected result**

tx-sender should success to update to FAILED and continue reading messages from kafka

**Logs**

tx-sender (codefi)

```
{"@timestamp":"2021-11-15T11:15:04Z","consumer":"c-0","job":"93996e51-bc11-441c-8418-3faec95ca578","log.level":"i │
│ nfo","message":"(use-cases.send-eth-tx): ethereum transaction job was sent successfully","schedule_uuid":"9ad32e5 │
│ c-79d5-4b36-b766-127d89e71ef4","tenant_id":"Q4fDt1s3jgFXqTTjXT7WhwYRnOY6ziE1:817bf907-7a37-4cf9-a50f-b2b807d8a37c │
│ "}                                                                                                                │
│ {"@timestamp":"2021-11-15T11:15:14Z","consumer":"c-0","job":"5c81ca66-e3ad-4c23-ad29-8cb3cd6d105f","log.level":"i │
│ nfo","message":"(use-cases.send-eth-tx): ethereum transaction job was sent successfully","schedule_uuid":"bb9a69e │
│ 6-49f3-4fa3-a82b-6cb9d4e08a22","tenant_id":"Q4fDt1s3jgFXqTTjXT7WhwYRnOY6ziE1:817bf907-7a37-4cf9-a50f-b2b807d8a37c │
│ "}                                                                                                                │
│ {"@timestamp":"2021-11-15T11:54:52Z","consumer":"c-0","error":"BE000@ethclient.rpc: code: -32603 - message: Backe │
│ nd connection failed","job":"a13acbed-587e-4bf6-ae6e-5654b69d35cc","log.level":"error","message":"(nonce-manager) │
│ : cannot retrieve fetch nonce from chain","schedule_uuid":"9f340543-7883-4fb2-ba80-22f50280dea8","tenant_id":"eph │
│ emeral1"}                                                                                                         │
│ {"@timestamp":"2021-11-15T11:54:52Z","log.level":"info","message":"(application): signal intercepted (ignored)"," │
│ sig":"broken pipe"}                                                                                               │
[...]

PR #747, opened 2021-11-23
#743 fix tx-sender crashes

ggarri:
## PR description
- Tx-sender should not exit when jobs are already in final status
- Ensure Tx-sender sends messages to tx-recover topic when there are persistent connectivity issues with RPC nodes

## Fixed Issue(s)
#743 

## Changelog


services/api/business/use-cases/jobs/update_job.go
```diff
@@ -206,7 +207,7 @@ func canUpdateStatus(nextStatus, status entities.JobStatus) bool {
 	case entities.StatusStored:
 		return status == entities.StatusStarted || status == entities.StatusRecovering
 	case entities.StatusFailed:
-		return status == entities.StatusStarted || status == entities.StatusRecovering || status == entities.StatusPending
+		return status == entities.StatusStarted || status == entities.StatusRecovering || status == entities.StatusPending || status == entities.StatusWarning || status == entities.StatusResending
```
6l20, line 210: This kind of statement is not easy to read, maybe we should consider having a State machine to make statuses management a bit clearer

services/tx-sender/integration-tests/ethereum_test.go
```diff
@@ -285,6 +285,54 @@ func (s *txSenderEthereumTestSuite) TestTxSender_Ethereum_Public() {
...
 	})
 
+	s.T().Run("should send envelope to tx-recover and don't retry if key-manager fails to sign and to update status to FAILED", func(t *testing.T) {
+		defer gock.Off()
+		wg := &multierror.Group{}
+
+		envelope := fakeEnvelope()
+
+		url := fmt.Sprintf("/ethereum/accounts/%s/sign-transaction", envelope.GetFromString())
+		gock.New(keyManagerURL).Post(url).Reply(http2.StatusUnauthorized).Status(422).
```
6l20, line 295: Use Http statuses instead of "magic numbers"

services/tx-sender/integration-tests/ethereum_test.go
```diff
@@ -285,6 +285,54 @@ func (s *txSenderEthereumTestSuite) TestTxSender_Ethereum_Public() {
...
+		if err != nil {
+			assert.Fail(t, err.Error())
+			return
+		}
+
+		err = waitTimeout(wg, time.Second*2)
+		assert.NoError(t, err)
+
+		assert.Equal(t, envelope.GetID(), retrievedEnvelope.GetID())
+		assert.NotEmpty(t, retrievedEnvelope.GetErrors()[0].Code)
```
6l20, line 332: We should also test the code value here

services/tx-sender/service/listener.go
```diff
@@ -23,6 +22,8 @@ import (
 
 const messageListenerComponent = "service.kafka-consumer"
 
+const errorProcessingMessage = "error processing message"
```
6l20, line 25: group the const together

services/tx-sender/service/listener.go
```diff
@@ -108,23 +109,44 @@ func (listener *MessageListener) consumeClaimLoop(ctx context.Context, session s
...
+				if serr != nil {
+					jlogger.WithError(serr).Error(errorProcessingMessage)
+					return serr
+				}
+			case err != nil:
+				curJob, serr := listener.jobClient.GetJob(ctx, job.UUID)
+				// IMPORTANT: Jobs can be updated in parallel to NEVER_MINED, MINED or FAILED, so that we should
+				// warning and ignore it in case job is in a final status
+				if serr == nil && entities.IsFinalJobStatus(curJob.Status) {
+					jlogger.WithError(err).Warn(errorProcessingMessage)
```
6l20, line 135: This one is always logged it seems

services/api/business/use-cases/jobs/update_job.go
ggarri: This is actually a state machine statemenent. What do you propose exactly to make it betteR?

services/tx-sender/integration-tests/ethereum_test.go
ggarri: Updated

services/tx-sender/service/listener.go
ggarri: What do you mean? it would be logged only in case job is already on a final status and that was the cause of the underline error

services/tx-sender/service/listener.go
ggarri: Done

services/api/business/use-cases/jobs/update_job.go
6l20: I meant a state machine using a state pattern, that would make things easier to read and handle. Sometimes when you start to have so many conditions, typically more than 3 in one single statement, it is a good sign that another design pattern is needed.

services/tx-sender/service/listener.go
6l20: ok, did not see first was warning only. my mistake

6l20 (approved):

Issue #767, opened 2021-12-08
Add in transaction receipt the contract name and contract tag


julien-marchand:
Add inside the transaction receipt, along with the decoded logs, the contract name, and contract tag that exactly match the ABI.

PR #773, opened 2021-12-14
767 contract search

ggarri:
## Fixed Issue(s)
#767 

## Changelog

- [x] I thought about the changelog and included a [changelog update if required](https://github.com/consensys/orchestrate/blob/master/CHANGELOG.md).


services/api/store/postgres/data-agents/contract.go
```diff
@@ -0,0 +1,82 @@
...
+`
+	_, err := agent.db.QueryOneContext(ctx, qContract, query, codeHash)
+	if err != nil {
+		return nil, pg.ParsePGError(err)
+	}
+
+	return parseContract(qContract), nil
+}
+
+func (agent *PGContract) FindOneByEventSigHash(ctx context.Context, sigHash string) (*entities.Contract, error) {
```
julien-marchand, line 54: I think this full flow should be removed. To me, there is quite a problem with FindOneByEventSigHash
if you look at OwnershipTransfered event see here: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.4.0/contracts/access/Ownable.sol many contracts can have this event so basically your just picking one arbitrarily and returning it. We should just look by codehash IMO

services/api/store/postgres/data-agents/contract.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func (agent *PGContract) FindOneByCodeHash(ctx context.Context, codeHash string) (*entities.Contract, error) {
+	qContract := &contractQuery{}
+	query := `
+SELECT a.abi, a.bytecode, a.deployed_bytecode, r.name as name, t.name as tag 
+FROM artifacts a
+INNER JOIN tags t ON (a.id = t.artifact_id)
+INNER JOIN repositories r ON (r.id = t.repository_id) 
+WHERE a.codehash = ?
```
julien-marchand, line 42: Shall we change the DB type as codehash is a fixed length?

pkg/toolkit/database/postgres/hook.go
```diff
@@ -9,7 +9,7 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-func handleError(err error) error {
+func ParsePGError(err error) error {
```
darioAnongba, line 12: needs to be exported?

pkg/types/api/account_request.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 type CreateAccountRequest struct {
-	Alias      string            `json:"alias" validate:"omitempty" example:"personal-account"`
+	Alias      string            `json:"alias" validate:"omitempty" example:"personal-account" ` // This is my alias description
```
darioAnongba, line 8: is that comment needed?

darioAnongba (approved):

pkg/toolkit/database/postgres/hook.go
ggarri: yes, since it is not used by the contract agent

pkg/types/api/account_request.go
ggarri: That was an example I was working out with byron from pliny. Undoing

services/api/store/postgres/data-agents/contract.go
ggarri: Removed in favor of looking by address

services/api/store/postgres/data-agents/contract.go
ggarri: I added indexes over `codehash` and `address` and according to my reach an exact match on TEXT fields are using an internal hash map, therefore it is fine to keep it as it is. Just in case of substring search it would be more optimal to have fixed length on columns

Issue #768, opened 2021-12-08
Create & sign using specific store


julien-marchand:
### Summary / User Story

As a user, I want to create accounts and sign using accounts using the non-default key store.

## Business Value

Adding support for this will allow a single Orchestrate stack to be connected to multiple vaults in different environments, thus avoiding to do the custody of the keys.

### Acceptance Criteria

- [ ] Add documentation examples

PR #777, opened 2021-12-15
#768 drafted logic for multi store support

ggarri:
## Fixed Issue(s)
#768 

## Changelog


6l20 (approved):

pkg/types/api/account_request.go
```diff
@@ -7,13 +7,15 @@ import (
 type CreateAccountRequest struct {
 	Alias      string            `json:"alias" validate:"omitempty" example:"personal-account" `
 	Chain      string            `json:"chain" validate:"omitempty" example:"besu"`
+	StoreID    string            `json:"storeID" validate:"omitempty" example:"qkmStoreID"`
```
darioAnongba, line 10: Why not just call it `Store json:"store"` ?

pkg/types/api/account_request.go
```diff
@@ -7,13 +7,15 @@ import (
...
 	Attributes map[string]string `json:"attributes,omitempty"`
 }
 
 type ImportAccountRequest struct {
 	Alias      string            `json:"alias" validate:"omitempty" example:"personal-account"`
 	Chain      string            `json:"chain" validate:"omitempty" example:"quorum"`
 	PrivateKey hexutil.Bytes     `json:"privateKey" validate:"required" example:"0x66232652FDFFD802B7252A456DBD8F3ECC0352BBDE76C23B40AFE8AEBD714E2D" swaggertype:"string"`
+	StoreID    string            `json:"storeID" validate:"omitempty" example:"qkmStoreID"`
```
darioAnongba, line 18: Why not just call it Store json:"store" ?

pkg/types/entities/account.go
```diff
@@ -14,6 +14,7 @@ type Account struct {
 	CompressedPublicKey hexutil.Bytes
 	TenantID            string
 	OwnerID             string
+	StoreID             string
```
darioAnongba, line 17: Why not just call it Store ?

services/api/business/use-cases/accounts/create_account.go
```diff
@@ -40,7 +40,7 @@ func NewCreateAccountUseCase(db store.DB, searchUC usecases.SearchAccountsUseCas
 		keyManagerClient: keyManagerClient,
 		fundAccountUC:    fundAccountUC,
 		logger:           log.NewLogger().SetComponent(createAccountComponent),
-		storeName:        qkm.GlobalStoreName(),
+		defaultStoreID:   qkm.GlobalStoreName(),
```
darioAnongba, line 43: The use case should not be aware of these defaults, it is the parent that will check if a `store` is defined in the request and fill it with the default if it's empty. When reaching the use case, the account MUST have a store defined

services/api/business/use-cases/accounts/create_account.go
```diff
@@ -64,10 +64,14 @@ func (uc *createAccountUseCase) Execute(ctx context.Context, account *entities.A
 		return nil, errors.AlreadyExistsError(errMsg).ExtendComponent(createAccountComponent)
 	}
 
+	if account.StoreID != "" {
```
darioAnongba, line 67: The business layer should not be aware of these defaults, it is the service layer that will check if a `store` is defined in the request and fill it with the default if it's empty. When reaching the use case, the account MUST have a store defined

services/api/store/postgres/migrations/22_add_account_storeID.go
```diff
@@ -0,0 +1,38 @@
+package migrations
+
+import (
+	"github.com/go-pg/migrations/v7"
+	log "github.com/sirupsen/logrus"
+)
+
+func upgradeAccountStoreID(db migrations.DB) error {
```
darioAnongba, line 8: If you already had accounts before v21.12. It means that you should first run the migrations, then import then accounts into the store. Are we missing something in the `imports` script?

services/api/service/formatters/accounts.go
```diff
@@ -14,13 +14,15 @@ func FormatCreateAccountRequest(req *api.CreateAccountRequest) *entities.Account
 	return &entities.Account{
 		Alias:      req.Alias,
 		Attributes: req.Attributes,
+		StoreID:    req.StoreID,
```
darioAnongba, line 17: this is where the default value should be set if empty in the request.

services/api/service/formatters/accounts.go
```diff
@@ -14,13 +14,15 @@ func FormatCreateAccountRequest(req *api.CreateAccountRequest) *entities.Account
...
 	}
 }
 
 func FormatImportAccountRequest(req *api.ImportAccountRequest) *entities.Account {
 	return &entities.Account{
 		Alias:      req.Alias,
 		Attributes: req.Attributes,
+		StoreID:    req.StoreID,
```
darioAnongba, line 25: this is where the default value should be set if empty in the request.

pkg/types/api/account_request.go
ggarri: Discussed an defined on storeID

services/api/business/use-cases/accounts/create_account.go
ggarri: After an Slack discussion it was decided to move on with this approach and create a tech debt ticket to apply same strategy for other default fields in the codebase such as priority or retryInterval

services/api/store/postgres/migrations/22_add_account_storeID.go
ggarri: Accounts having an empty StoreID will be using default StoreID.

services/api/business/use-cases/accounts/create_account.go
ggarri: Removed

services/api/service/formatters/accounts.go
ggarri: Applied change

services/api/service/formatters/accounts.go
ggarri: applied

Issue #791, opened 2022-01-04
Empty transaction fields in job


darioAnongba:
### Describe the bug
 
The `transaction` of the job is returned with all empty fields except the `gasPrice` field.

### Configuration

Orchestrate v21.12.0

### Steps to reproduce

1. Send a tx, get UUID
2. Wait for it to be mined
3. Query tx by UUID
4. Check the `jobs[0].transaction` field

**Actual result**

```
{
    "uuid": "1b950052-2724-4e43-ac4a-e3fb1107be64",
    "idempotencyKey": "",
    "chain": "geth",
    "params": {
        "from": "0x7e654d251da770a068413677967f6d3ea2fea9e4",
        "contractName": "Test",
        "contractTag": "v1.1.5"
    },
    "jobs": [
        {
            "uuid": "8f72dcc2-0bcb-42a2-a768-9dbc9b9e1d3f",
            "chainUUID": "d2eded04-cc55-4586-b120-376672fa68dc",
            "scheduleUUID": "1b950052-2724-4e43-ac4a-e3fb1107be64",
            "tenantID": "_",
            "transaction": {
                "gasPrice": "0x67209f1b",
                "createdAt": "2022-01-04T11:43:25.897912Z",
                "updatedAt": "2022-01-04T11:43:32.933941Z"
            },
            "labels": {
                "environmentName": "codefi-local-env-001",
                "serviceName": "Postman",
                "serviceUrl": "",
                "tenantId": "750037370"
            },
            "annotations": {
                "gasPricePolicy": {
                    "priority": "medium",
                    "retryPolicy": {}
                }
            },
            "status": "MINED",
            "type": "eth://ethereum/transaction",
            "createdAt": "2022-01-04T11:43:25.897912Z",
            "updatedAt": "2022-01-04T11:43:32.935762Z"
        }
    ],
    "createdAt": "2022-01-04T11:43:25.895719Z"
}
```

**Expected result**

Complete list of fields in the `transaction`

ggarri:
https://github.com/ConsenSys/orchestrate/pull/792/files

PR #792, opened 2022-01-04
Fix update not zero model values

ggarri:
## Fixed Issue(s)
#791 

## Changelog


pkg/toolkit/database/postgres/query.go
```diff
@@ -29,15 +29,15 @@ func InsertQuery(_ context.Context, q *orm.Query) error {
...
+func UpdateModel(ctx context.Context, db DB, models ...interface{}) error {
+	_, err := db.ModelContext(ctx, models...).WherePK().UpdateNotZero()
 	if err != nil {
 		return ParsePGError(err)
 	}
 	return nil
 }
 
-func UpdateNotZero(ctx context.Context, q *orm.Query) error {
+func Update(ctx context.Context, q *orm.Query) error {
```
darioAnongba, line 40: Why not just call `UpdateNotZero` instead of changing the function name?

services/api/store/postgres/data-agents/account.go
```diff
@@ -54,7 +54,7 @@ func (agent *PGAccount) Update(ctx context.Context, account *models.Account) err
 
 	account.UpdatedAt = time.Now().UTC()
 	agent.db.ModelContext(ctx, account)
-	err := pg.Update(ctx, agent.db, account)
+	err := pg.UpdateModel(ctx, agent.db, account)
```
darioAnongba, line 57: Why not just call `UpdateNotZero` ?

pkg/toolkit/database/postgres/query.go
ggarri: Because this name correlates better to the rest of query methods as you see follow:
```
func Delete(ctx context.Context, q *orm.Query) error {
	_, err := q.Context(ctx).Delete()
	if err != nil {
		return ParsePGError(err)
	}

	return nil
}

func Select(ctx context.Context, q *orm.Query) error {
	err := q.Context(ctx).Select()
	if err != nil {
		return ParsePGError(err)
	}

	return nil
}
```

services/api/store/postgres/data-agents/account.go
ggarri: answered above

PR #799, opened 2022-01-07
support for accesslog toggle

ggarri:
## Changelog


pkg/toolkit/app/http/router/dynamic/dynamic.go
```diff
@@ -49,15 +48,6 @@ func NewBuilder(staticCfg *traefikstatic.Configuration, epLogConfigs map[string]
 		logger:       log.NewLogger().SetComponent("router"),
 	}
 
-	for epName, logConfig := range epLogConfigs {
```
ggarri, line 52: It does not have any real usage in the code

pkg/toolkit/app/options.go
```diff
@@ -122,42 +121,6 @@ func LoggerMiddlewareOpt(midName string) Option {
 	}
 }
 
-func SwaggerOpt(specsFile string, middlewares ...string) Option {
```
ggarri, line 125: Broken feature

6l20 (approved):

PR #828, opened 2022-01-19
Duplicate account fix when importing 

darioAnongba:
Fix of old behavior when creating accounts.

**Old behavior**
When creating a new account, we were first creating it in the QKM and then checking that the account exists in Orchestrate. This logic is wrong for the following reasons:
- If creating a new account, it is statistically impossible to have a collision in accounts so checking that the account exists after a new one has been created if useless.
- If importing an account, checking that the account exists locally after it has been imported in the QKM is wrong because the account can exist in the QKM but not locally, meaning that we get a DependencyFailure instead of a 409 because the error happens on the QKM.

**Correct behavior**
- Check existence of the account locally only when we are importing and only call the QKM if the account does not exist. If the account exist, it MUST exist in the QKM as per design.


pkg/ethereum/account/account.go
```diff
@@ -3,60 +3,43 @@ package account
...
-	w.priv = priv
+	return &Account{
+		priv:    prv,
+		Address: crypto.PubkeyToAddress(prv.PublicKey),
+	}, nil
 }
 
-func (w *account) SetAddress(addr ethcommon.Address) {
-	w.address = addr
+func (a *Account) Priv() []byte {
```
ggarri, line 43: Is this method needed?

darioAnongba: The whole struct is not really needed but it's used for convenience.

services/api/business/use-cases/accounts/create_account.go
```diff
@@ -66,10 +65,27 @@ func (uc *createAccountUseCase) Execute(ctx context.Context, account *entities.A
...
+			logger.WithError(err).Error("invalid private key")
+			return nil, errors.InvalidParameterError(der.Error()).ExtendComponent(createAccountComponent)
+		}
+
+		_, der = uc.db.Account().FindOneByAddress(ctx, importedAccount.Address.Hex(), userInfo.AllowedTenants, userInfo.Username)
+		if der == nil {
+			errMsg := "account already exists"
+			logger.Error(errMsg)
+			return nil, errors.AlreadyExistsError(errMsg).ExtendComponent(createAccountComponent)
+		} else if !errors.IsNotFoundError(der) {
```
ggarri, line 82: No need for `else if` since you return on previous statement

services/api/business/use-cases/accounts/create_account.go
```diff
@@ -66,10 +65,27 @@ func (uc *createAccountUseCase) Execute(ctx context.Context, account *entities.A
...
 	if privateKey != nil {
-		resp, err = uc.keyManagerClient.ImportEthAccount(ctx, account.StoreID, &qkmtypes.ImportEthAccountRequest{
+		importedAccount, der := account.NewAccountFromPrivateKey(privateKey.String())
+		if der != nil {
+			logger.WithError(err).Error("invalid private key")
+			return nil, errors.InvalidParameterError(der.Error()).ExtendComponent(createAccountComponent)
+		}
+
+		_, der = uc.db.Account().FindOneByAddress(ctx, importedAccount.Address.Hex(), userInfo.AllowedTenants, userInfo.Username)
+		if der == nil {
```
ggarri, line 78: IMO this logic is hard to read I would prefer something more explicit like 

```
existingAcc, der = uc.db.Account().FindOneByAddress(ctx, importedAccount.Address.Hex(), userInfo.AllowedTenants, userInfo.Username)
		if existingAcc != nil {
                        errMsg := "account already exists"
			logger.Error(errMsg)
			return nil, errors.AlreadyExistsError(errMsg).ExtendComponent(createAccountComponent)
		} 

              if der != nil && !errors.IsNotFoundError(der) {
                       errMsg := "failed to get account"
			logger.WithError(der).Error(errMsg)
			return nil, errors.FromError(der).ExtendComponent(createAccountComponent)
		}
```

ggarri (approved):

services/api/business/use-cases/accounts/create_account.go
darioAnongba: applied

services/api/business/use-cases/accounts/create_account.go
darioAnongba: applied

Issue #833, opened 2022-01-20
Implement transaction cancellation and retry endpoints


julien-marchand:
### Summary / User Story

User transactions might get stuck in tx-pool due to nonce or gasprice issues. In order to help users with this particular scenario Orchestrate should provide two endpoints
1. `{tx-id}/cancel` : Send replacement transaction with the same nonce, empty data, and at least 10% more gas than the previous transaction so it encourages miners to take if first and replace previous queue tx
2. `{tx-id}/speed-up`: Send the same transaction with the same nonce but with higher gas price

## Business Value
Users cannot use Orchestrate to unblock the queue of pending txs in the pool

### Acceptance Criteria

- [ ] Implement a cancel tx endpoint 
- [ ] Implement a speed-up tx endpoint

PR #837, opened 2022-01-24
Entrypoints to speed-up or call-off a pending transaction

ggarri:
## Fixed Issue(s)
#833 

## Changelog


services/api/business/use-cases/jobs.go
```diff
@@ -55,3 +57,7 @@ type UpdateChildrenUseCase interface {
 type ResendJobTxUseCase interface {
 	Execute(ctx context.Context, jobUUID string, userInfo *multitenancy.UserInfo) error
 }
+
+type RetryJobTxUseCase interface {
```
darioAnongba, line 61: What's the difference between "resend" and "retry"? I don't remember

services/api/business/use-cases/jobs/retry_job_tx.go
```diff
@@ -0,0 +1,88 @@
...
+		startJobUC:    startJobUC,
+		logger:        log.NewLogger().SetComponent(retryJobTxComponent),
+	}
+}
+
+// Execute sends a job to the Kafka topic
+func (uc *retryJobTxUseCase) Execute(ctx context.Context, jobUUID string, gasIncrement float64, txData hexutil.Bytes, userInfo *multitenancy.UserInfo) error {
+	ctx = log.WithFields(ctx, log.Field("job", jobUUID))
+	logger := uc.logger.WithContext(ctx)
+	logger.Debug("resending job transaction")
```
darioAnongba, line 41: "retrying" ?

services/api/business/use-cases/jobs/retry_job_tx.go
```diff
@@ -0,0 +1,88 @@
...
+	retriedJob, err := uc.createJobTxUC.Execute(ctx, job, userInfo)
+	if err != nil {
+		return errors.FromError(err).ExtendComponent(retryJobTxComponent)
+	}
+
+	if err = uc.startJobUC.Execute(ctx, retriedJob.UUID, userInfo); err != nil {
+		return errors.FromError(err).ExtendComponent(retryJobTxComponent)
+	}
+
+	logger.WithField("job", retriedJob.UUID).Info("job was retried successfully")
```
darioAnongba, line 86: "job retried successfully", you can remove the "was"

services/api/business/use-cases/jobs/retry_job_tx.go
```diff
@@ -0,0 +1,88 @@
...
+	if job.Status != entities.StatusPending {
+		errMessage := "cannot retry job transaction at the current status"
+		logger.WithField("status", job.Status).Error(errMessage)
+		return errors.InvalidStateError(errMessage)
+	}
+
+	job.InternalData.ParentJobUUID = jobUUID
+	job.Transaction.Data = txData
+	if job.Transaction.TransactionType == entities.LegacyTxType {
+		if job.Transaction.GasPrice == nil {
```
darioAnongba, line 58: that's impossible right? If the tx is in PENDING, it MUST have a gas price right?

services/api/business/use-cases/jobs/retry_job_tx.go
```diff
@@ -0,0 +1,88 @@
...
+			errMsg := "transaction has no gas price set"
+			logger.Error(errMsg)
+			return errors.InvalidParameterError(errMsg).ExtendComponent(retryJobTxComponent)
+		}
+		gasPrice := job.Transaction.GasPrice.ToInt()
+		txGasPrice := gasPrice.Mul(gasPrice, big.NewInt(10)).Div(gasPrice, big.NewInt(100))
+		job.Transaction.GasPrice = utils.ToPtr(hexutil.Big(*txGasPrice)).(*hexutil.Big)
+	} else {
+		if job.Transaction.GasTipCap == nil {
```
darioAnongba, line 67: that's impossible right? If the tx is in PENDING, it MUST have a gas tip cap right?

services/api/business/use-cases/transactions/call_off_tx.go
```diff
@@ -0,0 +1,65 @@
...
+	logger := uc.logger.WithContext(ctx)
+	logger.Debug("calling off pending transaction")
+
+	tx, err := uc.getTxUC.Execute(ctx, scheduleUUID, userInfo)
+	if err != nil {
+		return nil, errors.FromError(err).ExtendComponent(sendTxComponent)
+	}
+
+	if tx.Params.Protocol != "" {
+		errMsg := "cancel transaction is not supported"
```
darioAnongba, line 47: ".. is not supported for private transactions" ?
I don't think this needs to be returned as "FeatureNotSupported" because I'm not sure this feature makes sense for private txs. It's more like an `InvalidParameter` but I'm not sure what error code should be returned here because this will never be supported.

services/api/business/use-cases/transactions/call_off_tx.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/consensys/orchestrate/pkg/toolkit/app/log"
+	"github.com/consensys/orchestrate/pkg/toolkit/app/multitenancy"
+	"github.com/consensys/orchestrate/pkg/types/entities"
+	usecases "github.com/consensys/orchestrate/services/api/business/use-cases"
+	"github.com/consensys/orchestrate/services/api/store"
+)
+
+const callOffTxComponent = "use-cases.call-off-tx"
+
+// callOffTxUseCase is a use case to get a transaction request
```
darioAnongba, line 16: better to remove all these useless comments

services/api/business/use-cases/transactions/speed_up_tx.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/consensys/orchestrate/pkg/toolkit/app/log"
+	"github.com/consensys/orchestrate/pkg/toolkit/app/multitenancy"
+	"github.com/consensys/orchestrate/pkg/types/entities"
+	usecases "github.com/consensys/orchestrate/services/api/business/use-cases"
+	"github.com/consensys/orchestrate/services/api/store"
+)
+
+const speedUpTxComponent = "use-cases.speed-up-tx"
+
+// speedUpTxUseCase is a use case to get a transaction request
```
darioAnongba, line 16: better to remove all these useless comments

services/api/business/use-cases/transactions/speed_up_tx.go
```diff
@@ -0,0 +1,66 @@
...
+	)
+	logger := uc.logger.WithContext(ctx)
+	logger.Debug("speeding up transaction")
+
+	tx, err := uc.getTxUC.Execute(ctx, scheduleUUID, userInfo)
+	if err != nil {
+		return nil, errors.FromError(err).ExtendComponent(sendTxComponent)
+	}
+
+	if tx.Params.Protocol != "" {
```
darioAnongba, line 47: same remark. Seeping up transactions doesn't make sense for private txs because they do not compete with each other on the mempool

services/api/service/controllers/transactions.go
```diff
@@ -269,3 +275,64 @@ func (c *TransactionsController) search(rw http.ResponseWriter, request *http.Re
...
+	if err != nil {
+		httputil.WriteHTTPErrorResponse(rw, err)
+		return
+	}
+
+	_ = json.NewEncoder(rw).Encode(formatters.FormatTxResponse(txRequest))
+}
+
+// @Summary Call of transaction
+// @Description Call of transaction  by sending empty data transaction with same nonce and additional 10% more gas
```
darioAnongba, line 316: double space here. You can try `swag fmt` : https://github.com/swaggo/swag#the-swag-formatter

services/api/business/use-cases/jobs.go
ggarri: Resend: It is sending the very same tx to the node, it involves just a new log entry 
Retry: Retry a job with tx changes such as gasPrice and Data, t involves a new job entry

services/api/business/use-cases/jobs/retry_job_tx.go
ggarri: Indeed I added for the cases where network gas is 0, but checking the execution flow, it should be set to 0x0, it is not nil. Removing this

services/api/business/use-cases/jobs/retry_job_tx.go
ggarri: As above, I was expecting nil for 0 gas nets but the value is actually 0x0 so I am removing this now

services/api/business/use-cases/transactions/call_off_tx.go
ggarri: I also hesitates in this, I can use invalid parameter but as you mention it does not appear to fit at 100%

services/api/business/use-cases/transactions/call_off_tx.go
ggarri: Totally agree, removed

services/api/business/use-cases/jobs/retry_job_tx.go
ggarri: good catch

services/api/business/use-cases/jobs/retry_job_tx.go
ggarri: done

services/api/business/use-cases/transactions/speed_up_tx.go
ggarri: removed

services/api/business/use-cases/transactions/speed_up_tx.go
ggarri: Yes, that is why this condition was added. Is there any proposed change here?

services/api/service/controllers/transactions.go
ggarri: added in the version latest version, cool!

services/api/business/use-cases/jobs.go
darioAnongba: ok

services/api/business/use-cases/transactions/speed_up_tx.go
darioAnongba: no, same comment on the error message + type

Issue #803, opened 2022-01-10
[tech debt] Remove entities from API types and Models


ggarri:
### Summary / User Story

Entities are considered as internal DTOs for our application and they should never be exposed externally, therefore they should never be used as part of user-facing types as API or DB models. 

## Business Value
Having Entities integrated as part of our API types or models create a high level of coupling between different layers which lead to side issues like the one addressed in this ticket https://github.com/ConsenSys/orchestrate/issues/800

### Acceptance Criteria

- [ ] No entities are part of API types
- [ ] No entities are part of Model types
- [ ] Only Entities are used to exchange messages between layers

PR #842, opened 2022-01-28
803 tech debt entities

ggarri:
## PR description
- Removed usage of models from business layer

## Fixed Issue(s)
#803 

## Changelog


src/api/business/use-cases/accounts/search_identity_test.go
```diff
@@ -41,7 +41,7 @@ func TestSearchAccounts_Execute(t *testing.T) {
 		resp, err := usecase.Execute(ctx, filter, userInfo)
 
 		assert.NoError(t, err)
-		assert.Equal(t, parsers2.NewAccountEntityFromModels(acc), resp[0])
+		assert.Equal(t, parsers2.NewAccountEntity(acc), resp[0])
```
darioAnongba, line 44: not sure this is needed anymore? The `NewAccountEntity` function

src/api/business/use-cases/chains/delete_chain_test.go
```diff
@@ -70,7 +70,7 @@ func TestDeleteChain_Execute(t *testing.T) {
 	t.Run("should fail with same error if delete private tx manager fails", func(t *testing.T) {
 		expectedErr := errors.NotFoundError("error")
 		chain := testdata.FakeChain()
-		chainModel := parsers.NewChainModelFromEntity(chain)
+		chainModel := parsers.NewChainModel(chain)
```
darioAnongba, line 73: we should not have the concept of models in the business layer no right? (applies for all unit tests)

src/api/business/use-cases/jobs/create_job.go
```diff
@@ -60,30 +58,29 @@ func (uc *createJobUseCase) Execute(ctx context.Context, job *entities.Job, user
 		}
 	}
 
-	schedule, err := uc.db.Schedule().FindOneByUUID(ctx, job.ScheduleUUID, userInfo.AllowedTenants, userInfo.Username)
+	scheduleID, err := uc.db.Schedule().GetIDByUUID(ctx, job.ScheduleUUID, userInfo.AllowedTenants, userInfo.Username)
```
darioAnongba, line 61: I don't think the ID should be used outside of the data layer. This ID is mostly internal to postgres, it's not something that should be exposed because the business layer is agnostic of that. Let say we switch from postgres to MongoDB, the business layer should not be affected

src/api/business/use-cases/jobs/create_job.go
```diff
@@ -60,30 +58,29 @@ func (uc *createJobUseCase) Execute(ctx context.Context, job *entities.Job, user
...
+	job.Status = entities.StatusCreated
+	job.Logs = append(job.Logs, &entities.Log{
 		Status: entities.StatusCreated,
 	})
-	jobModel.Schedule = schedule
-
-	if err = uc.db.Transaction().Insert(ctx, jobModel.Transaction); err != nil {
+	
+	var txID int
+	if txID, err = uc.db.Transaction().Insert(ctx, job.Transaction); err != nil {
```
darioAnongba, line 74: same remark returning the txID here. The fact that the ID is an `int` is not something the business layer can know.

src/api/business/use-cases/accounts/search_identity_test.go
ggarri: you are right, I didn't work on the UTs yet. Waiting for the overall agreement on the impl before doing it

src/api/business/use-cases/chains/delete_chain_test.go
ggarri: you are right, I didn't work on the UTs yet. Waiting for the overall agreement on the impl before doing it

src/api/business/use-cases/jobs/create_job.go
ggarri: Remove every knowledge of models IDs from business layer and data-agent interfaces

src/api/business/use-cases/jobs/create_job.go
ggarri: Remove every knowledge of models IDs from business layer and data-agent interfaces

src/api/business/use-cases/accounts/search_identity_test.go
ggarri: done

src/api/business/use-cases/chains/delete_chain_test.go
ggarri: done

darioAnongba (approved):

PR #843, opened 2022-01-28
Redis infra refactor

darioAnongba:
Implement redis infra implementation:

- Add REDIS_IDLE_TIMEOUT ENV var
- Adds REDIS_MAX_IDLE ENV var
- Removes TLS certs as strings but now expects paths to certificates
- Move flags to `cmd`

Little refactor of the interface


scripts/coverage.sh
```diff
@@ -4,7 +4,7 @@
 set -Eeu
 
 # Ignore generated & testutils files
-cat $1 | grep -Fv -e ".pb.go" -e ".pb.gw.go" -e "/tests" -e "/testutils" -e "/integration-test" -e "mock/" -e "mocks/" -e "/integration-test" -e "/migrations" > "$1.tmp"
+cat $1 | grep -Fv -e ".pb.go" -e ".pb.gw.go" -e "/tests" -e "/testutils" -e "/integration-test" -e "mock/" -e "mocks/" -e "/integration-test" -e "/migrations" -e "/infra" > "$1.tmp"
```
ggarri, line 7: this only remove covered values from the reports. If you want to ignore `/infra` from sonar you have to do it in `./sonar-properties` at the root folder

src/infra/redis/redigo/client.go
```diff
@@ -0,0 +1,130 @@
+package redigo
```
ggarri, line 1: isn't it possible to maintain historical changes of this file? it looks like you rewrote it all

src/infra/redis/redigo/config.go
```diff
@@ -0,0 +1,91 @@
+package redigo
```
ggarri, line 1: same above keeping historical record of the file changes

scripts/coverage.sh
ggarri: Anyway this only remove covered values from the reports. If you want to ignore `/infra` from sonar you have to do it in `./sonar-properties` at the root folder

src/tx-sender/tx-sender/nonce/manager/manager.go
```diff
@@ -0,0 +1,191 @@
+package manager
```
ggarri, line 1: keep history of this file

scripts/coverage.sh
darioAnongba: removed, I thought it was only unit tests

src/tx-sender/tx-sender/nonce/manager/manager.go
darioAnongba: I think I made a mistake while refactoring the folder structure. I followed the pattern "interface at the root and individual folders for each concrete implementation". So the old file was modified into an interface and the new file contains all the old changes. I'm not sure it's possible to do the opposite because Git will think the file has been modified.

src/infra/redis/redigo/config.go
darioAnongba: I think I made a mistake while refactoring the folder structure. I followed the pattern "interface at the root and individual folders for each concrete implementation". So the old file was modified into an interface and the new file contains all the old changes. I'm not sure it's possible to do the opposite because Git will think the file has been modified.

src/infra/redis/redigo/client.go
darioAnongba: I think I made a mistake while refactoring the folder structure. I followed the pattern "interface at the root and individual folders for each concrete implementation". So the old file was modified into an interface and the new file contains all the old changes. I'm not sure it's possible to do the opposite because Git will think the file has been modified.

ggarri (approved):

PR #850, opened 2022-02-01
803 tech debt entities

ggarri:
## PR description
- Removed usage of entities on API responses
- Removed complex types from API responses
- Removed duplicated and redundant types such as ETHTramsactionParams
- 
## Fixed Issue(s)
#803 

## Changelog


pkg/types/privacy/types.go
```diff
@@ -0,0 +1,8 @@
+package privacy
+
+type PrivateETHTransactionParams struct {
```
darioAnongba, line 3: Why is there an exception for this type such that it needs its own package?

pkg/crypto/ethereum/signing.go
```diff
@@ -33,7 +33,7 @@ func SignQuorumPrivateTransaction(tx *quorumtypes.Transaction, privKey *ecdsa.Pr
 	return decodedSignature, nil
 }
 
-func SignEEATransaction(tx *types.Transaction, privateArgs *entities.PrivateETHTransactionParams, chainID *big.Int, privKey *ecdsa.PrivateKey) ([]byte, error) {
+func SignEEATransaction(tx *types.Transaction, privateArgs *privacy.PrivateETHTransactionParams, chainID *big.Int, privKey *ecdsa.PrivateKey) ([]byte, error) {
```
darioAnongba, line 36: This package is useless, these functions should not be part of a package but be part of the use case. There is only one place in the codebase where `SignEEATransaction` should be used. Same for all functions of the `crypto/ethereum` package.

src/api/service/formatters/accounts.go
```diff
@@ -50,9 +50,9 @@ func FormatAccountResponse(iden *entities.Account) *api.AccountResponse {
 	return &api.AccountResponse{
 		Alias:               iden.Alias,
 		Attributes:          iden.Attributes,
-		Address:             iden.Address,
-		PublicKey:           iden.PublicKey,
-		CompressedPublicKey: iden.CompressedPublicKey,
+		Address:             iden.Address.String(),
```
darioAnongba, line 53: rather use `Hex()` I think. `Hex()` does `String()` internally but that's internal to go-ethereum

src/api/service/formatters/chains.go
```diff
@@ -20,19 +20,30 @@ func FormatChainResponse(chain *entities.Chain) *types.ChainResponse {
...
-		PrivateTxManager:          chain.PrivateTxManager,
+		PrivateTxManager:          FormatPrivateTxManagerResponse(chain.PrivateTxManager),
 		Labels:                    chain.Labels,
 		CreatedAt:                 chain.CreatedAt,
 		UpdatedAt:                 chain.UpdatedAt,
 	}
 }
 
+func FormatPrivateTxManagerResponse(txManager *entities.PrivateTxManager) *types.PrivateTxManagerResponse {
+	if txManager == nil {
```
darioAnongba, line 37: I'm normally against functions that do this check. If the `txManager` is `nil`, then this function should not be called from the parent. We don't want any "logic" here.

src/api/service/formatters/contracts.go
```diff
@@ -1,14 +1,13 @@
 package formatters
```
darioAnongba, line 1: for me the best would be to completely remove the `formatters` folder and to implement the service layer types in the same way as the models. Example:

```
type XRequest struct {
   ...
}

function (r *XRequest) ToEntity() *entitites.X {
   return &entities.X{...}
}

type XResponse struct {
   ...
}

function NewXResponse(x *entities.X) *XResponse {
   return &XResponse{....}
}
```

So we can get rid of the `parsers` and `formatters` folders everywhere.

src/api/service/formatters/eth_transaction.go
```diff
@@ -0,0 +1,147 @@
...
+		PrivateFrom:     tx.PrivateFrom,
+		PrivateFor:      tx.PrivateFor,
+		MandatoryFor:    tx.MandatoryFor,
+		PrivacyGroupID:  tx.PrivacyGroupID,
+		PrivacyFlag:     entities.PrivacyFlag(tx.PrivacyFlag),
+		CreatedAt:       tx.CreatedAt,
+		UpdatedAt:       tx.UpdatedAt,
+	}
+
+	if tx.Hash != "" {
```
darioAnongba, line 94: Same here with the pointer issue

src/api/service/formatters/transactions.go
```diff
@@ -20,24 +20,26 @@ func FormatSendTxRequest(sendTxRequest *types.SendTransactionRequest, idempotenc
...
-			To:              sendTxRequest.Params.To,
-			Value:           sendTxRequest.Params.Value,
-			GasPrice:        sendTxRequest.Params.GasPrice,
-			Gas:             sendTxRequest.Params.Gas,
-			GasFeeCap:       sendTxRequest.Params.GasFeeCap,
-			GasTipCap:       sendTxRequest.Params.GasTipCap,
-			AccessList:      sendTxRequest.Params.AccessList,
-			TransactionType: sendTxRequest.Params.TransactionType,
+		Params: &entities.TxRequestParams{
+			ETHTransaction: &entities.ETHTransaction{
```
darioAnongba, line 24: You seem to have added one "level" for nothing as all your fields are inside that new level now

src/api/service/formatters/transactions.go
```diff
@@ -59,24 +61,26 @@ func FormatDeployContractRequest(deployRequest *types.DeployContractRequest, ide
...
-			ContractName:    deployRequest.Params.ContractName,
-			ContractTag:     deployRequest.Params.ContractTag,
-			Protocol:        deployRequest.Params.Protocol,
-			PrivateFrom:     deployRequest.Params.PrivateFrom,
-			PrivateFor:      deployRequest.Params.PrivateFor,
-			MandatoryFor:    deployRequest.Params.MandatoryFor,
-			PrivacyFlag:     entities.PrivacyFlag(deployRequest.Params.PrivacyFlag),
-			PrivacyGroupID:  deployRequest.Params.PrivacyGroupID,
+		Params: &entities.TxRequestParams{
+			ETHTransaction: &entities.ETHTransaction{
```
darioAnongba, line 65: same comment on "level"

src/api/service/formatters/transactions.go
```diff
@@ -111,16 +117,18 @@ func FormatTransferRequest(transferRequest *types.TransferRequest, idempotencyKe
...
-			GasPrice:        transferRequest.Params.GasPrice,
-			Gas:             transferRequest.Params.Gas,
+		Params: &entities.TxRequestParams{
+			ETHTransaction: &entities.ETHTransaction{
+				From:            &transferRequest.Params.From,
+				To:              &transferRequest.Params.To,
+				GasFeeCap:       transferRequest.Params.GasFeeCap,
+				GasTipCap:       transferRequest.Params.GasTipCap,
+				AccessList:      transferRequest.Params.AccessList,
+				TransactionType: entities.TransactionType(transferRequest.Params.TransactionType),
```
darioAnongba, line 127: This is an example of what I meant before. `TransactionType` does not exist in `ETHTransaction`

src/api/service/formatters/eth_transaction.go
```diff
@@ -0,0 +1,147 @@
...
+		PrivateFrom:     tx.PrivateFrom,
+		PrivateFor:      tx.PrivateFor,
+		MandatoryFor:    tx.MandatoryFor,
+		PrivacyGroupID:  tx.PrivacyGroupID,
+		PrivacyFlag:     int(tx.PrivacyFlag),
+		CreatedAt:       tx.CreatedAt,
+		UpdatedAt:       tx.UpdatedAt,
+	}
+
+	if tx.Hash != nil {
```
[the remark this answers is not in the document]
ggarri: tx.Hash can be empty if its value was not computed by the sender yet. The fact of need pointers and check whether or not they are nil it is a casuistic of our type formatted I cannot find out how to go around unless you define a type of TX per state and you check the state on the parser to make sure which values must be filled up and which are not

pkg/types/privacy/types.go
ggarri: I've check and it is only use by Sender UCs, so I will move this there

src/api/business/use-cases/accounts/fund_account.go
```diff
@@ -64,10 +64,12 @@ func (uc *fundAccountUseCase) Execute(ctx context.Context, account *entities.Acc
 	txRequest := &entities.TxRequest{
 		IdempotencyKey: utils.RandString(16),
 		ChainName:      chains[0].Name,
-		Params: &entities.ETHTransactionParams{
-			From:  &faucet.CreditorAccount,
-			To:    &account.Address,
-			Value: &faucet.Amount,
+		Params: &entities.TxRequestParams{
+			ETHTransaction: &entities.ETHTransaction{
```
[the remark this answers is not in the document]
ggarri: Question is, do we need to differentiate them? Are there any business constraints that forces us to maintain to types just to split what comes from the user and what does not? IMO there is not value in keeping them independent since our business layer manages them in an interchangeable way. 

In addition at the service layer, we already have defined a ETHTransactionRequest and ETHTransactionResponse as the types to exchange with the API users.

src/api/service/formatters/accounts.go
ggarri: After all both of them are doing the same
```
func (a Address) Hex() string {
	return string(a.checksumHex())
}

// String implements fmt.Stringer.
func (a Address) String() string {
	return a.Hex()
}
```

src/api/service/formatters/chains.go
ggarri: Indeed I agree and also noticed but since it was like that I left it. I will do so in my next version

src/api/service/formatters/eth_transaction.go
ggarri: Same response as it was added above. Let resolve this in favor of the other one to keep up the discussion there

src/api/service/formatters/contracts.go
ggarri: This looks like what I proposed at the beginning to start working in Orchestrate and we moved to formatters and parsers for reasons I cannot remember. IMO this approach is much cleaner, so if you agree I am gonna work on that too

src/api/service/formatters/transactions.go
ggarri: Since entities are purely internal there is not constrains in the way we build them up, I can split ETHParams and rest of params into two separate properties, I did this way so it can be easily modified across the code with less changes. No a big deal if you prefer to have a flatter format

src/api/service/formatters/transactions.go
ggarri: Lets keep discussion on the dedupe comment. Resolving

src/api/service/formatters/transactions.go
ggarri: what it is the point here?

src/api/service/formatters/contracts.go
ggarri: Creating a ticket to apply this

darioAnongba (approved):

src/api/service/formatters/contracts.go
ggarri: https://app.zenhub.com/workspaces/tanuki-5ea70772b186e10067f57842/issues/consensys/orchestrate/851

pkg/types/privacy/types.go
ggarri: done

pkg/crypto/ethereum/signing.go
ggarri: done

PR #891, opened 2022-03-03
Infra postgres refactor

darioAnongba:
## PR description

- Refactor postgres infra
- Removes Tx and DB interfaces
- Aggregates PrivateTxManager and Chains
- Modifies DB TXs logic
- Removes parsers
- Refactors array of pointers to simple arrays
- Removes `STORE_TYPE` ENV var
- Moves config to `cmd`


cmd/flags/postgres.go
```diff
@@ -239,93 +226,28 @@ const (
...
-}
-
-func Copy(opts *pg.Options) *pg.Options {
-	if opts == nil {
-		return nil
+		SSLMode:           vipr.GetString(DBTLSSSLModeViperKey),
+		TLSCert:           vipr.GetString(DBTLSCertViperKey),
+		TLSKey:            vipr.GetString(DBTLSKeyViperKey),
+		TLSCA:             vipr.GetString(DBTLSCAViperKey),
+		ApplicationName:   "api.store",
```
ggarri, line 251: I believe the PG settings should not contain the app destination. I would suggest to attach this value one the config it is passed to the application requiring the usage of it

src/api/config.go
```diff
@@ -47,7 +47,7 @@ type Config struct {
 func NewConfig(vipr *viper.Viper) *Config {
 	return &Config{
 		App:          app.NewConfig(vipr),
-		Store:        store.NewConfig(vipr),
+		Postgres:     flags.NewPGConfig(vipr),
 		Multitenancy: viper.GetBool(multitenancy.EnabledViperKey),
```
ggarri, line 51: it should be ` vipr.GetBool()` insead, otherway we won't grab the values from the cmd params loaded in the vipr object

src/api/exported.go
```diff
@@ -25,16 +24,19 @@ func New(ctx context.Context) (*app.App, error) {
 		return nil, err
 	}
 
+	postgresClient, err := gopg.New(cfg.Postgres)
```
ggarri, line 27: It might not be the scope of this PR but I would suggest to modify the New method to accept the a cfg instead of initializing it here.

```
func New(ctx context.Context, cfg *Config) (*app.App, error) {

```

src/api/store/postgres/store.go
```diff
@@ -1,72 +1,86 @@
...
+
+func (s *PGStore) Chain() store.ChainAgent {
+	return s.chain
+}
 
-	return nil
+func (s PGStore) RunInTransaction(ctx context.Context, persist func(a store.DB) error) error {
+	return s.client.RunInTransaction(ctx, func(dbTx postgres.Client) error {
+		s.client = dbTx
+		return persist(&s)
```
ggarri, line 84: If I understand it properly, this won't work. The Agents under `s` will remain having the old client and not the new one, you would need to do instead
```
func (s PGStore) RunInTransaction(ctx context.Context, persist func(a store.DB) error) error {
	return s.client.RunInTransaction(ctx, func(dbTx postgres.Client) error {
		s2 := New(dbTx)
		return persist(s2)
	})
}
```

or better for performance enhancement
```
func (s PGStore) RunInTransaction(ctx context.Context, persist func(a store.DB) error) error {
	return s.client.RunInTransaction(ctx, func(dbTx postgres.Client) error {
		if s.client.AlreadyADbTx() {
			return persist(s)
		}
		return persist(New(dbTx))
	})
}
```

where you would need to add a Internface method to validate whether or not the dbClient is already under dbTx

cmd/api/migrate.go
```diff
@@ -11,33 +11,31 @@ import (
...
-			db = pg.Connect(opts)
+
+			db = pg.Connect(pgOptions)
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return migrate(db)
 		},
-		PersistentPostRun: func(cmd *cobra.Command, args []string) {
-			err := db.Close()
```
ggarri, line 32: Where are you performing the close db connection now?

src/api/business/builder/transactions.go
```diff
@@ -34,8 +34,8 @@ func newTransactionUseCases(
 		sendTransaction:         sendTxUC,
 		getTransaction:          getTransactionUC,
 		searchTransactions:      transactions.NewSearchTransactionsUseCase(db, getTransactionUC),
```
ggarri, line 36: Is there any reason to inject the `db` object instead of passing the required DbAgent directly here? `db.TransactionRequest()`

src/api/business/use-cases/contracts/register_contract.go
```diff
@@ -119,3 +107,46 @@ func getIndexedCount(event *abi.Event) (indexedInputCount uint) {
 
 	return indexedInputCount
 }
+
+// TODO: Remove this function as parsing the events from the ABI should not be done on Orchestrate as we do not have control on how the events are represented in the ABI
+
+func ParseEvents(data string) (map[string]string, error) {
```
ggarri, line 113: this method can be private

src/api/store/models/transaction_request.go
```diff
@@ -18,3 +19,43 @@ type TransactionRequest struct {
...
+	return &TransactionRequest{
+		IdempotencyKey: txRequest.IdempotencyKey,
+		ChainName:      txRequest.ChainName,
+		RequestHash:    requestHash,
+		Params:         txRequest.Params,
+		CreatedAt:      txRequest.CreatedAt,
+	}
+}
+
+func NewTxRequests(txRequests []TransactionRequest) []entities.TxRequest {
```
ggarri, line 33: You forgot to migrate all the unit tests around parser implementations.

src/api/store/postgres/chain.go
```diff
@@ -0,0 +1,238 @@
...
+func NewPGChain(client postgres.Client) *PGChain {
+	return &PGChain{
+		client: client,
+		logger: log.NewLogger().SetComponent("data-agents.chain"),
+	}
+}
+
+func (agent *PGChain) Insert(ctx context.Context, chain *entities.Chain) error {
+	return agent.client.RunInTransaction(ctx, func(c postgres.Client) error {
+		model := models.NewChain(chain)
```
ggarri, line 35: WE should minimize the time within a DBTX, therefore I would suggest to move model crafting can happen outside of this callback

src/api/store/postgres/chain.go
```diff
@@ -0,0 +1,238 @@
...
+			if err != nil {
+				errMessage := "failed to insert private tx manager"
+				agent.logger.WithContext(ctx).WithError(err).Error(errMessage)
+				return errors.FromError(err).SetMessage(errMessage)
+			}
+
+			return nil
+		}
+
+		utils.CopyPtr(model.ToEntity(), chain)
```
ggarri, line 66: Move it out of the DBTX

src/api/store/postgres/chain.go
```diff
@@ -0,0 +1,238 @@
...
+		q = gopg.WhereAllowedOwner(q, "", ownerID)
+
+		err = c.Update(q)
+		if err != nil {
+			errMessage := "failed to update chain"
+			agent.logger.WithContext(ctx).WithError(err).Error(errMessage)
+			return errors.FromError(err).SetMessage(errMessage)
+		}
+
+		utils.CopyPtr(model.ToEntity(), chain)
```
ggarri, line 207: Move it out

ggarri (changes requested):
Additional comments:
- We lost of the code coverage around parsers
- We lost historical records on the data-agents

src/api/business/builder/transactions.go
darioAnongba: done like that in other Use cases, can be modified globally in a PR

src/api/business/use-cases/contracts/register_contract.go
darioAnongba: resolved

src/api/store/models/transaction_request.go
darioAnongba: WIP

src/api/store/postgres/chain.go
darioAnongba: resolved

src/api/store/postgres/chain.go
darioAnongba: resolved

src/api/store/postgres/chain.go
darioAnongba: resolved

cmd/api/migrate.go
darioAnongba: resolved

ggarri (approved):

Issue #904, opened 2022-03-16
Remove usage of protobuf ethereum types


ggarri:
## Background
Orchestrate uses auto-generate protobuf objects to encapsulate receipt and transaction information retrieved from Blockchain. 

## Acceptance criteria
- [ ] Remove usage of protobuf object from ethereum client component
- [ ] Only use protobuf for external communications using kafka

## Technical notes
- Use native go-ethereum types to retrieve data from blockchain network
- In case of go-quorum specific types import them from go-quorum library
- In case of eea specific types, create mirror Go types

Issue #884, opened 2022-03-02
Use go-web3 to decode logs


ggarri:
Rewrite current in-house logic to decode logs in `chain-listener` by go-web3 in order to reduce maintenance effort on our side

## Acceptance criteria

- [ ] Check that structs can be decoded

darioAnongba:
Before going to go-web3 we could try to see if go-ethereum can handle that using: https://github.com/ethereum/go-ethereum/blob/master/accounts/abi/bind/base.go

PR #905, opened 2022-03-16
228 Centralize kafka tx response within API service 

ggarri:
## PR description
- Only API service sends notifications to the clients
- Notifications are crafted under API job update flow
- Removed unnecessary update validation
- Refactor job data-agent to aggregate dbtx logic
- Fixed bug on nested RunInTransaction 

## Changelog


CHANGELOG.md
```diff
@@ -1,11 +1,11 @@
 # Orchestrate Release Notes
 
-## v22.2.0 (Unreleased)
+## v22.5.0 (Unreleased)
```
darioAnongba, line 3: How can we already have a 22.5 ? That seems impossible...

pkg/types/ethereum/receipt.go
```diff
@@ -5,12 +5,29 @@ import (
...
 	"errors"
 
+	"github.com/consensys/orchestrate/pkg/utils"
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
 )
 
+type internalReceipt struct {
```
darioAnongba, line 15: is this temporary?

src/api/business/builder/contracts.go
```diff
@@ -7,23 +7,26 @@ import (
...
 type contractUseCases struct {
-	GetContractsCatalogUC      usecases.GetContractsCatalogUseCase
+	getContractsCatalogUC      usecases.GetContractsCatalogUseCase
 	getContractEvents          usecases.GetContractEventsUseCase
 	getContractTags            usecases.GetContractTagsUseCase
 	registerContractDeployment usecases.RegisterContractDeploymentUseCase
 	registerContractUC         usecases.RegisterContractUseCase
 	getContractUC              usecases.GetContractUseCase
 	searchContractUC           usecases.SearchContractUseCase
+	decodeLogUC                usecases.DecodeEventLogUseCase
```
darioAnongba, line 17: is this a standalone UC ? Seems more like this UC is used in an other UC

src/api/business/use-cases/contracts/decode_event_log.go
```diff
@@ -0,0 +1,125 @@
+package contracts
```
darioAnongba, line 1: good to check if we can avoid doing this ourselves and use a package.

src/api/business/use-cases/jobs.go
```diff
@@ -46,9 +44,8 @@ type UpdateJobUseCase interface {
 	Execute(ctx context.Context, jobEntity *entities.Job, nextStatus entities.JobStatus, logMessage string, userInfo *multitenancy.UserInfo) (*entities.Job, error)
 }
 
-type UpdateChildrenUseCase interface {
-	Execute(ctx context.Context, jobUUID, parentJobUUID string, nextStatus entities.JobStatus, userInfo *multitenancy.UserInfo) error
-	WithDBTransaction(dbtx store.DB) UpdateChildrenUseCase
+type UpdateJobStatusUseCase interface {
```
darioAnongba, line 47: IMO we should separate this into smaller use cases because they will grow in complexity really quick when we implement MQs. Can be done later.

src/api/business/use-cases/jobs/notify_job_update.go
```diff
@@ -0,0 +1,182 @@
...
+
+	// Send message
+	partition, offset, err := uc.kafkaProducer.SendMessage(msg)
+	if err != nil {
+		errMsg := "could not produce kafka message"
+		logger.WithError(err).Error(errMsg)
+		return errors.KafkaConnectionError(errMsg)
+	}
+
+	logger.WithField("partition", partition).
```
darioAnongba, line 136: This is not an INFO log

src/api/business/use-cases/jobs/start_job.go
```diff
@@ -63,15 +58,14 @@ func (uc *startJobUseCase) Execute(ctx context.Context, jobUUID string, userInfo
 		return errors.FromError(err).ExtendComponent(startJobComponent)
 	}
 
-	partition, offset, err := SendJobMessage(job, uc.kafkaProducer, uc.topicsCfg.Sender)
+	err = uc.notifyJobUpdate.Execute(ctx, job, "")
```
darioAnongba, line 61: code smell since you need an empty string here?

src/api/business/use-cases/jobs/notify_job_update.go
```diff
@@ -0,0 +1,182 @@
...
+		db:               db,
+		kafkaProducer:    kafkaProducer,
+		topicsCfg:        topicsCfg,
+		searchContractUC: searchContractUC,
+		decodeLogUC:      decodeLogUC,
+		logger:           log.NewLogger().SetComponent(notifyJobUpdateComponent),
+	}
+}
+
+func (uc *notifyJobUpdateUseCase) Execute(ctx context.Context, job *entities.Job, notifyErrMsg string) error {
```
darioAnongba, line 51: For me this is not a use case, as proven by the switch you have. This is just a `Notifier` component that should have multiple specific functions depending on what you want to achieve instead of a use case with `Execute` that does all the logic in switches and having optional params.

src/api/business/use-cases/jobs/resend_job_tx.go
```diff
@@ -50,12 +46,13 @@ func (uc *resendJobTxUseCase) Execute(ctx context.Context, jobUUID string, userI
 		return errors.InvalidStateError(errMessage)
 	}
 
-	partition, offset, err := SendJobMessage(job, uc.kafkaProducer, uc.topicsCfg.Sender)
+	job.Status = entities.StatusStarted
+	err = uc.notifyJobUpdate.Execute(ctx, job, "")
```
darioAnongba, line 50: same issue with notifier not being a use case

CHANGELOG.md
ggarri: haha, that was a rough estimation on when we will release this. But we can go back to 22.3.0 (WIP)

pkg/types/ethereum/receipt.go
ggarri: Yes, I added this ticket to make sure we get rid of this before next release https://github.com/ConsenSys/orchestrate/issues/904

src/api/business/builder/contracts.go
ggarri: This UC is currently only use on jobs.SendNotifiticationUC but it is expected to be called by other users cases once we implement smart contract subscriptions

src/api/business/use-cases/contracts/decode_event_log.go
ggarri: We have a ticket to deal with this independently https://github.com/ConsenSys/orchestrate/issues/884

src/api/business/use-cases/jobs.go
ggarri: IMO once the entire DBTX complexity is moved into the data-layer this complexity of this UC would be deducted significantly

src/api/business/use-cases/jobs/notify_job_update.go
ggarri: done

src/api/business/use-cases/jobs/start_job.go
ggarri: I could have created on single notification UC per job status so I remove this but IMO since 98% of the code it is the same it is not a good idea to just duplicate use cases for this kind of scenarios

src/api/business/use-cases/jobs/notify_job_update.go
ggarri: Same priciple could be applied to every "childUseCase" as we used to call it, or any UseCase which is not trigger by users directly, but our current arch does not follow that practice and modifying that would involve a large amount of work when the benefit might not directly clear. Since this use-case, as others, has a clear scoped work and responsibility which it runs independent I would rather keep this approach to prevent us to start a new large paradigm refactor of our codebase.

src/api/business/use-cases/jobs/resend_job_tx.go
ggarri: Resolving in favor of the above open discussion

src/api/business/builder/contracts.go
ggarri: I could have created on single notification UC per job status so I remove this but IMO since 98% of the code it is the same it is not a good idea to just duplicate use cases for this kind of scenarios

src/api/store/postgres/job.go
```diff
@@ -30,52 +30,103 @@ func NewPGJob(client postgres.Client) *PGJob {
...
+			jobLogModel.JobID = &curJobModel.ID
+			jobLogModel.UUID = uuid.Must(uuid.NewV4()).String()
+			jobLogModel.CreatedAt = jobModel.UpdatedAt
+			err = dbtx.ModelContext(ctx, jobLogModel).Insert()
+			if err != nil {
+				return err
+			}
+		}
+
+		if job.Transaction != nil {
```
darioAnongba, line 112: Here we use `job.Transaction` instead of passing the transaction in the params of the function like `jobLog`.

src/api/service/types/job.go
```diff
@@ -22,6 +23,7 @@ type UpdateJobRequest struct {
 	Labels      map[string]string      `json:"labels,omitempty"` // List of custom labels.
 	Annotations *Annotations           `json:"annotations,omitempty"`
 	Transaction *ETHTransactionRequest `json:"transaction,omitempty"`
+	Receipt     *ethereum.Receipt      `json:"receipt,omitempty"`
```
darioAnongba, line 26: Why does `Receipt` come from `pkg` It's an entity like all the others no?

src/api/service/formatters/jobs.go
```diff
@@ -59,9 +59,10 @@ func FormatJobCreateRequest(request *types.CreateJobRequest) *entities.Job {
 	return job
 }
 
-func FormatJobUpdateRequest(request *types.UpdateJobRequest) *entities.Job {
+func FormatJobUpdateRequest(request *types.UpdateJobRequest) (*entities.Job, error) {
```
darioAnongba, line 62: Formatters shouldn't fail normally as it's passed validation

src/api/service/controllers/jobs.go
```diff
@@ -214,7 +214,12 @@ func (c *JobsController) update(rw http.ResponseWriter, request *http.Request) {
 		return
 	}
 
-	job := formatters.FormatJobUpdateRequest(jobRequest)
+	job, err := formatters.FormatJobUpdateRequest(jobRequest)
+	if err != nil {
```
darioAnongba, line 218: should not fail

src/api/service/controllers/contracts.go
```diff
@@ -196,7 +196,7 @@ func (c *ContractsController) setCodeHash(rw http.ResponseWriter, request *http.
 // @Failure      400        {object}  httputil.ErrorResponse                     "Invalid request"
 // @Failure      404        {object}  httputil.ErrorResponse                     "Events not found"
 // @Failure      500        {object}  httputil.ErrorResponse                     "Internal server error"
-// @Router       /contracts/accounts/{chain_id}/{address}/events [get]
+// @Router       /contracts/accounts/{chain_uuid}/{address}/events [get]
```
darioAnongba, line 199: Not sure I understand the change here. Do we need to use chain UUID ? Since events are tied to a blockchain so a chainID

src/api/business/use-cases/notifications/notify_failed_job.go
```diff
@@ -0,0 +1,64 @@
...
+		Key:   sarama.StringEncoder(job.ChainUUID),
+		Value: sarama.ByteEncoder(b),
+	})
+	if err != nil {
+		errMsg := "could not send kafka notification"
+		logger.WithError(err).Error(errMsg)
+		return errors.KafkaConnectionError(errMsg)
+	}
+
+	logger.Info("FAILED job notification was sent successfully")
```
darioAnongba, line 61: I think  we should only log INFO when the state of the app changes or on the actual main use case. Here it's DEBUG

src/api/business/use-cases/jobs/resend_job_tx.go
```diff
@@ -44,18 +43,12 @@ func (uc *resendJobTxUseCase) Execute(ctx context.Context, jobUUID string, userI
...
+	err = sendEnvelope(uc.kafkaProducer, uc.topicsCfg.Sender, job)
 	if err != nil {
-		logger.WithError(err).Error("failed to send job message")
-		return errors.FromError(err).ExtendComponent(resendJobTxComponent)
+		logger.WithError(err).Error("failed to send resend job envelope")
+		return err
 	}
 
-	logger.WithField("partition", partition).WithField("offset", offset).Info("job resend successfully")
+	logger.Info("job resend successfully")
```
darioAnongba, line 52: "transaction resent successfully" ?

src/api/business/use-cases/contracts/decode_event_log.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+const decodeLogsComponent = "use-cases.decode-logs"
+
+type decodeEventLogUseCase struct {
+	getContractEventsUC usecases.GetContractEventsUseCase
+	logger              *log.Logger
+}
+
+func NewDecodeEventLogUseCase(getContractEventsUC usecases.GetContractEventsUseCase) usecases.DecodeEventLogUseCase {
```
darioAnongba, line 28: IMO this could be done using `go-ethereum` because I'm pretty sure decoding logs doesn't work with latest solidity features.

src/api/service/formatters/jobs.go
ggarri: actually that was not needed, undoing

src/api/service/controllers/contracts.go
ggarri: You are right chainID is the actual value needed, but since our APIs understand chainUUID and in order to be consistent. In before this in our codebase we are fetching the chain first in other to make this call with `chainID` what seems not very natural

src/api/business/use-cases/notifications/notify_failed_job.go
ggarri: Moving to DEBUG

src/api/business/use-cases/jobs/start_job.go
```diff
@@ -47,64 +47,31 @@ func (uc *startJobUseCase) Execute(ctx context.Context, jobUUID string, userInfo
...
-		return nil
-	})
+	err = sendEnvelope(uc.kafkaProducer, uc.topicsCfg.Sender, curJob)
 	if err != nil {
+		logger.WithError(err).Error("failed to send start job envelope")
 		return err
 	}
 
-	uc.addMetrics(job.UpdatedAt.Sub(prevUpdatedAt), prevStatus, status, job.ChainUUID)
+	logger.Info("job started successfully")
```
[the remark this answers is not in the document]
ggarri: IIRC we didn't the statement of adding INFO when the state of the APP changes, in this case the job has changed to STARTED, therefore INFO sounds correct

src/api/business/use-cases/jobs/resend_job_tx.go
ggarri: sounds better

src/api/business/use-cases/contracts/decode_event_log.go
ggarri: Since we have a dedicated ticket for it I would rather wait for it to deal with this topic https://github.com/ConsenSys/orchestrate/issues/884

src/api/store/postgres/job.go
ggarri: It was done like that because jobLog is built internally by Orchestrate as part of the execution flow in meanwhile transaction object is built out of the input from the user in the formatters `FormatJobUpdateRequest`.

darioAnongba (approved):

PR #911, opened 2022-03-21
Forward JWT to tx-sender

darioAnongba:
Forward JWT to tx-sender


src/api/business/use-cases/jobs/utils.go
```diff
@@ -5,13 +5,15 @@ import (
...
+	"github.com/consensys/orchestrate/pkg/toolkit/app/multitenancy"
 	"github.com/consensys/orchestrate/src/entities"
 )
 
-func SendJobMessage(job *entities.Job, kafkaProducer sarama.SyncProducer, topic string) (partition int32, offset int64, err error) {
+func SendJobMessage(job *entities.Job, kafkaProducer sarama.SyncProducer, topic string, userInfo *multitenancy.UserInfo) (partition int32, offset int64, err error) {
 	txEnvelope := job.TxEnvelope(map[string]string{
-		authutils.TenantIDHeader: job.TenantID,
-		authutils.UsernameHeader: job.OwnerID,
+		authutils.TenantIDHeader:      job.TenantID,
```
ggarri, line 14: Since you injected the userInfo, I believe it makes more sense to attach the values for tenantID and ownerID from it

src/tx-sender/service/listener.go
```diff
@@ -108,15 +111,25 @@ func (listener *MessageListener) consumeClaimLoop(ctx context.Context, session s
 
 			jlogger := logger.WithField("job", evlp.GetJobUUID()).WithField("schedule", evlp.GetScheduleUUID())
 			job := entities.NewJobFromEnvelope(evlp)
-			err = listener.processEnvelope(log.With(ctx, jlogger), evlp, job)
+			newCtx := log.With(ctx, jlogger)
+
+			if evlp.Headers[authutils.AuthorizationHeader] != "" {
```
ggarri, line 116: Use the helper function
```
func WithAuthorization(ctx context.Context, authorization string) context.Context {
	return context.WithValue(ctx, authorizationKey, authorization)
}
```
and since the header attach already check for not empty values you can remove the if as well here
```
func (t *AuthHeadersTransport) RoundTrip(req *http.Request) (*http.Response, error) {
	if authutils.GetAuthorizationHeader(req) == "" {
		authutils.AddAuthorizationHeaderValue(req, t.auth)
	}

	return t.T.RoundTrip(req)
}
```

darioAnongba: done

ggarri (approved):

src/api/business/use-cases/jobs/utils.go
darioAnongba: done

PR #923, opened 2022-03-25
Contracts improvements

darioAnongba:
Improvements on the registration of contracts:
- A contract is overwritten when tag and name are the same
- Duplicate artifacts in DB
- Improvements on formatters
- Separation between update and register logic


src/api/business/use-cases/contracts/register_contract.go
```diff
@@ -37,22 +37,32 @@ func (uc *registerContractUseCase) Execute(ctx context.Context, contract *entiti
...
+	contract.RawABI, err = getABICompacted(contract.RawABI)
+	if err != nil {
+		errMessage := "failed to parse contract ABI"
+		logger.WithError(err).Error(errMessage)
+		return errors.InvalidParameterError(errMessage).ExtendComponent(registerContractComponent)
+	}
 	contract.CodeHash = crypto.Keccak256(contract.DeployedBytecode)
 
-	events, err := getEvents(&contract.ABI, contract.DeployedBytecode, crypto.Keccak256Hash(contract.DeployedBytecode), abiRaw)
+	events, err := getEvents(&contract.ABI, contract.DeployedBytecode, crypto.Keccak256Hash(contract.DeployedBytecode), contract.RawABI)
```
6l20, line 53: Here we should reuse the contract.CodeHash value computed above.
And as this function uses 3 contracts parameters, maybe we should consider passing contract as a parameter instead.

src/api/business/use-cases/contracts/register_contract.go
```diff
@@ -37,22 +37,32 @@ func (uc *registerContractUseCase) Execute(ctx context.Context, contract *entiti
...
-		logger.WithError(err).Error("failed to parse contract ABI")
-		return errors.FromError(err).ExtendComponent(registerContractComponent)
+		errMessage := "failed to parse contract events from ABI"
+		logger.WithError(err).Error(errMessage)
+		return errors.InvalidParameterError(errMessage).ExtendComponent(registerContractComponent)
 	}
 
-	// @TODO Improve duplicate inserts when `DeployedBytecode` and `Name` and `Tag` already exists
-	err = uc.db.Contract().Register(ctx, contract)
+	// If the same contract already exists, we update it with the new content
```
6l20, line 60: That comment only applies to line 64

6l20 (approved):

PR #935, opened 2022-03-30
EventStream model

darioAnongba:
Event stream model and creation use case. Only webhooks supported for now.


src/api/store/postgres/migrations/26_create_event_streams_table.go
```diff
@@ -0,0 +1,55 @@
...
+	log.Debug("Creating event streams table...")
+
+	_, err := db.Exec(`
+CREATE TABLE event_streams (
+	id SERIAL PRIMARY KEY,
+    uuid UUID NOT NULL,
+    tenant_id TEXT NOT NULL,
+    owner_id TEXT,
+    name TEXT,
+    chain TEXT NOT NULL,
```
ggarri, line 18: I though we decoded on doing tenant_id+chain unique, making chain nullable and allowing users to define a single stream for all its chains

src/api/store/postgres/migrations/26_create_event_streams_table.go
```diff
@@ -0,0 +1,55 @@
...
+    channel TEXT NOT NULL,
+    status TEXT NOT NULL,
+	specs JSONB NOT NULL,
+    labels JSON,
+	created_at TIMESTAMPTZ DEFAULT (now() at time zone 'utc') NOT NULL, 
+	updated_at TIMESTAMPTZ DEFAULT (now() at time zone 'utc') NOT NULL
+);
+
+CREATE UNIQUE INDEX event_streams_unique_name_idx ON event_streams (tenant_id, name) WHERE name IS NOT NULL;
```
ggarri, line 27:
````
CREATE UNIQUE INDEX event_streams_unique_chain_idx ON event_streams (tenant_id, chain) WHERE name IS NOT NULL;
```

darioAnongba: the name of the event stream should be unique per tenant so we need an UNIQUE on (name, tenant) as well no ?

src/api/store/postgres/migrations/26_create_event_streams_table.go
darioAnongba: I think calling the chain chain and set it as TEXT is better because we might change that to be the the  chain name or the chainID at some point and we won't need a migration.

src/api/store/postgres/migrations/26_create_event_streams_table.go
darioAnongba: addressed

src/api/store/postgres/migrations/26_create_event_streams_table.go
darioAnongba: addressed, called it chain_uuid and type UUID

PR #938, opened 2022-03-31
Feature/webhook notification

darioAnongba:
Webhook implementation:
- Infra component
- Crafting of the Notification
- Sending of the JSON response (mined + failed)
- Find event stream in DB


src/api/business/use-cases/event_streams/notify_transaction.go
```diff
@@ -0,0 +1,119 @@
...
+func (uc *notifyTransactionUseCase) Execute(ctx context.Context, job *entities.Job, errStr string, userInfo *multitenancy.UserInfo) error {
+	ctx = log.WithFields(ctx, log.Field("job_uuid", job.UUID), log.Field("job_status", job.Status))
+	logger := uc.logger.WithContext(ctx)
+
+	eventStream, err := uc.db.EventStream().FindOneByTenantAndChain(ctx, job.TenantID, job.ChainUUID, userInfo.AllowedTenants, userInfo.Username)
+	if err != nil {
+		return errors.FromError(err).ExtendComponent(notifyTransactionComponent)
+	}
+
+	if eventStream != nil && eventStream.Status == entities.EventStreamStatusLive {
```
ggarri, line 50: MINOR: I would rather simplify this code by doing early exit conditions such as follow

```
if eventStream != nil {
 return nil
}

if eventStream.Status != entities.EventStreamStatusLive  {
  return nil
}

// Rest of our code

File: src/infra/kafka/kafka.go [hunk 2]
```diff
@@ -17,7 +16,6 @@ type Client interface {
 
 type Producer interface {
 	SendJobMessage(topic string, job *entities.Job, userInfo *multitenancy.UserInfo) error
```
ggarri, line 18: You can add a // TODO here to remember to move this away from the `kafka` folder

ggarri (approved):

src/infra/kafka/kafka.go
darioAnongba: done

PR #965, opened 2022-04-28
Feature/push notification service

darioAnongba:
First iteration of the push notification service. Works fine but these are the missing features:
- Formatting of TxResponse and NotificationResponse
- Storage of Notification in DB + migration
- Acknowledgment of Notification in DB
- Retry policy to count the number of retries before suspending the event stream
- Unit tests
- Integration tests


src/api/store/models/event_stream.go
```diff
@@ -42,16 +42,15 @@ func NewEventStream(eventStream *entities.EventStream) *EventStream {
...
-		switch eventStream.Channel {
-		case entities.EventStreamChannelWebhook:
-			es.Specs = &EventStreamSpecs{
-				WebHook: eventStream.WebHook(),
-			}
-		case entities.EventStreamChannelKafka:
-			es.Specs = &EventStreamSpecs{
-				Kafka: eventStream.Kafka(),
-			}
+	if eventStream.Kafka != nil {
```
ggarri, line 45: We should do the split using the `switch eventStream.Channel` so the stream type and the specs match. Otherwise we are using a "side" conditional which is not coherent with the rest of the codebase

src/infra/ethclient/rpc/ethereum.go
```diff
@@ -441,7 +441,7 @@ func (ec *Client) SendTransaction(ctx context.Context, endpoint string, args *ty
 
 // SendRawPrivateTransaction send a raw transaction to an Ethereum node supporting EEA extension
 func (ec *Client) SendRawPrivateTransaction(ctx context.Context, endpoint string, raw hexutil.Bytes) (ethcommon.Hash, error) {
-	// Send a raw signed transactions using EEA extension method
+	// SendNotificationResponse a raw signed transactions using EEA extension method
```
ggarri, line 444: Invalid auto replacement

src/infra/messenger/kafka/producer.go
```diff
@@ -61,13 +66,64 @@ func (p *Client) SendJobMessage(topic string, job *entities.Job, partitionKey st
...
+		msg.Key = sarama.StringEncoder(partitionKey)
+	}
+
+	b, err := json.Marshal(notif)
+	if err != nil {
+		return err
+	}
+	msg.Value = sarama.ByteEncoder(b)
+
+	if userInfo.AuthMode == multitenancy.AuthMethodJWT {
```
ggarri, line 90: Why do we need userInfo in the notification message? Since notification it is trigger from a tx-listener request I believe userInfo will always be internal admin users.

src/notifier/notifier/use-cases/notifications/create-tx.go
```diff
@@ -0,0 +1,55 @@
...
+func NewCreateTransactionUseCase() usecases.CreateTxNotificationUseCase {
+	return &createTxUseCase{
+		logger: log.NewLogger().SetComponent(createTxComponent),
+	}
+}
+
+func (uc *createTxUseCase) Execute(ctx context.Context, job *entities.Job, errStr string) *entities.Notification {
+
+	notif := &entities.Notification{
+		SourceUUID: job.ScheduleUUID,
```
ggarri, line 29: We are missing the SoruceType, right? so that when we implement subscription we can differentiate the flow to trigger

src/notifier/notifier/use-cases/notifications/send.go
```diff
@@ -0,0 +1,56 @@
...
+		logger:          log.NewLogger().SetComponent(sendComponent),
+	}
+}
+
+func (uc *sendUseCase) Execute(ctx context.Context, notif *entities.Notification, eventStream *entities.EventStream) error {
+	logger := uc.logger.WithContext(log.WithFields(ctx, log.Field("notification", notif.UUID), log.Field("event_stream", eventStream.UUID)))
+
+	if eventStream.Status == entities.EventStreamStatusSuspend {
+		logger.Warn("event stream is suspended")
+		return nil
```
ggarri, line 34: Since use-case does not perform the expected action of sending the notification, I would rather expect and error here

src/notifier/notifier/use-cases/notifications/send.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	if err != nil {
+		errMessage := "failed to send notification to the specified event stream"
+		logger.WithError(err).Error(errMessage)
+		return errors.DependencyFailureError(errMessage)
+	}
+
+	// TODO(dario): Acknowledge notification
+
+	logger.Info("transaction notification sent successfully")
```
ggarri, line 54: We are missing the update of Notification to `SENT`, right?

ggarri (approved):

src/infra/messenger/kafka/producer.go
darioAnongba: done

src/infra/ethclient/rpc/ethereum_test.go
```diff
@@ -607,7 +608,7 @@ func TestSendTransaction(t *testing.T) {
 	// Test 1 with Error
 	ctx := testutils.NewContext(fmt.Errorf("test-error"), 0, nil)
 	_, err := ec.SendTransaction(ctx, "test-endpoint", &types.SendTxArgs{})
-	assert.Error(t, err, "#1 Send should  error")
+	assert.Error(t, err, "#1 SendNotificationResponse should  error")
```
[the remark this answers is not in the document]
darioAnongba: done

src/notifier/notifier/use-cases/notifications/create-tx.go
darioAnongba: done

src/notifier/notifier/use-cases/notifications/send.go
darioAnongba: discussed

src/notifier/notifier/use-cases/notifications/send.go
darioAnongba: discussed, next PR

Issue #937, opened 2022-03-31
Support subcriptions for contract events


ggarri:
## User story
- `user1` subscribes to `0xcontractaddrOne` from block `1` 
- `user1` receives notifications on the intended eventStream from block `1` to `latest`
- `user1` receives a new notification every time a new event happens on `0xcontractaddrOne`

## Requirements
- Notify “at-least-one” to user on subscribed events
- Minimize the amount of double notifications
- Optmize the number of RPC calls
- Compatible with tx-listener in HA mode

## Tech docs
- [API design](https://docs.google.com/document/d/19fFt5keREfTRCHYmRVWM9nQYW9deV7w_ZUwcKe_JNk8/edit#heading=h.3wtdxfjstbsy)
- [Listener execution flow](https://docs.google.com/document/d/1moY20PYiSF4xcpdoiNv2eLJwbtEWmcJR8QS_6eYl5IM/edit#heading=h.wrru038rtvr)

PR #973, opened 2022-05-02
937 subscription api

ggarri:
## PR description
- New `/subscription` API
- DB layer implemented
- Use Cases implemented
- Implemented API listener service 
- Small refactor on tx-listener service
- Small refactor on notifier service
- Small refactor infra/messager/producer
- Decouple API and Notifier service initialization
- Message SDK wrapper

## Fixed Issue(s)
#937

## Documentation

## Changelog


src/api/store/models/event_stream.go
```diff
@@ -44,16 +44,12 @@ func NewEventStream(eventStream *entities.EventStream) *EventStream {
 
 	switch eventStream.Channel {
 	case entities.EventStreamChannelKafka:
-		if eventStream.Kafka != nil {
-			es.Specs = &EventStreamSpecs{
-				Kafka: eventStream.Kafka,
-			}
+		es.Specs = &EventStreamSpecs{
```
darioAnongba, line 47: I think this is wrong because you will break the update when the specs are undefined

src/api/store/models/event_stream.go
```diff
@@ -83,13 +79,11 @@ func (e *EventStream) ToEntity() *entities.EventStream {
...
 	}
 
-	if e.Specs != nil {
-		switch es.Channel {
-		case entities.EventStreamChannelWebhook:
-			es.Webhook = e.Specs.WebHook
-		case entities.EventStreamChannelKafka:
-			es.Kafka = e.Specs.Kafka
-		}
+	switch es.Channel {
```
darioAnongba, line 82: Same here, you are breaking the update

src/api/store/models/subscription.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/consensys/orchestrate/src/entities"
+	ethcommon "github.com/ethereum/go-ethereum/common"
+)
+
+type Subscription struct {
+	tableName struct{} `pg:"subscriptions"` // nolint:unused,structcheck // reason
+
+	ID              int `pg:"alias:id"`
+	UUID            string
+	Address         string
```
darioAnongba, line 15: Rather call it ContractAddress

src/api/store/models/subscription.go
```diff
@@ -0,0 +1,67 @@
...
+	ContractName    string
```
darioAnongba, line 16: Why would you need contractName and contractTag? It's clear that this is duplicated no?

src/api/store/postgres/migrations/28_create_subscrition_table.go
```diff
@@ -0,0 +1,54 @@
...
+
+func createSubscriptionsTable(db migrations.DB) error {
+	log.Debug("Creating subscriptions table...")
+
+	_, err := db.Exec(`
+CREATE TABLE subscriptions (
+	id SERIAL PRIMARY KEY,
+    uuid UUID NOT NULL,
+	address TEXT,
+    tenant_id TEXT NOT NULL,
```
darioAnongba, line 16: tenant and owner not needed

src/entities/event_stream.go
```diff
@@ -17,6 +17,9 @@ var (
 	EventStreamStatusSuspend EventStreamStatus = "SUSPEND"
 )
 
+const WildcardChainUUID = "00000000-0000-0000-0000-000000000000"
```
darioAnongba, line 20: That's weird...

src/infra/messenger/kafka/consumer.go
```diff
@@ -114,20 +108,42 @@ func (cl *Consumer) consumeClaimLoop(ctx context.Context, session sarama.Consume
 
 			logger.WithField("timestamp", msg.Timestamp).Trace("message consumed")
 
-			decodedMsg, err := cl.handler.DecodeMessage(msg)
+			req := &types.ConsumerRequestMessage{}
```
darioAnongba, line 111: isn't ConsumerRequestMessage under same folder?

darioAnongba (approved):

src/api/store/models/subscription.go
```diff
@@ -0,0 +1,67 @@
...
+	ContractTag     string
+	EventStreamUUID string    `pg:"alias:event_stream_uuid"`
+	ChainUUID       string    `pg:"alias:chain_uuid"`
+	TenantID        string    `pg:"alias:tenant_id"`
```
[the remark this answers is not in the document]
ggarri: I see your point, but if we don't attach this information here we won't be able to search subscriptions per user(tenant+username). We cannot use eventStream, since streams can be reused across different tenants in case of using `_` to create it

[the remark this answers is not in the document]
ggarri: done

src/api/store/postgres/migrations/28_create_subscrition_table.go
ggarri: needed for filtering by tenant_id and username

src/entities/event_stream.go
ggarri: agree, but since we used chain_uuid column as UUID type, I needed a value which converts from Wildcard. Alternative it is to use chain_uuid as TEXT and use `*` directly. What do you prefer?

src/infra/messenger/kafka/consumer.go
ggarri: moved to the same folder

PR #1002, opened 2022-05-16
Performance testing

6l20:
## PR description

- Automate performance testing
- Discard private-tx testing

## Documentation

## Changelog


.circleci/config.yml
```diff
@@ -300,6 +300,25 @@ jobs:
             KAFKA_CONSUMER_GROUP_NAME: "stress"
           command: ./build/bin/test stress
 
+  run-perf:
```
ggarri, line 303: avoid no standardized acronyms so it is easy to read

Makefile
```diff
@@ -114,7 +114,7 @@ lint-tools: ## Install linting tools
 tools: lint-tools ## Install test tools
 	go install github.com/golang/mock/mockgen@v1.4.3
 	go install github.com/swaggo/swag/cmd/swag@v1.7.8
-	go install github.com/tsenart/vegeta
+	go get -u github.com/tsenart/vegeta
```
ggarri, line 117: `go get` install dependencies locally on the project. In this case we want vegeta to be installed globally

6l20: go install fails with
`no required module provides package github.com/tsenart/vegeta`
And go get adds dependencies indeed.

ggarri: Then use precompile ones https://github.com/tsenart/vegeta/releases

PR #1003, opened 2022-05-16
readiness url as param

6l20:
## PR description

- let sonar run on main also
- readiness url is set in qa and staging contexts (at org level)

## Documentation

## Changelog


.circleci/config.yml
```diff
@@ -511,6 +512,7 @@ workflows:
             - run-qa-e2e
       - run-e2e:
           name: run-staging-e2e
+          readiness-url: staging.orchestrate.ops.consensys.net
```
darioAnongba, line 515: No need to define a new parameter, the variable already exists in the context, use ${READINESS_URL}

.circleci/config.yml
```diff
@@ -491,6 +491,7 @@ workflows:
               only: /^v.*/
       - run-e2e:
           name: run-qa-e2e
+          readiness-url: qa.orchestrate.ops.consensys.net
```
darioAnongba, line 494: No need to define a new parameter, the variable already exists in the context, use ${READINESS_URL}

.circleci/config.yml
6l20: Ok, good

darioAnongba (approved):

PR #1005, opened 2022-05-16
Connection error when decoding logs

darioAnongba:
Connection error when decoding logs


services/tx-listener/session/ethereum/hooks/kafka/hook.go
```diff
@@ -246,36 +249,7 @@ func (hk *Hook) decodeReceipt(ctx context.Context, c *dynamic.Chain, receipt *ty
 		l.DecodedData = mapping
 		l.Event = GetAbi(event)
 
-		logger.WithField("receipt_log", fmt.Sprintf("%v", mapping)).
-			Debug("log decoded")
-	}
-
-	// It will only execute in cae of external txs
-	if receipt.ContractName == "" && contractAddress != nil {
```
ggarri, line 254: Please, revert this till we can confirm this is not in use by any client atm

6l20 (approved):

PR #1020, opened 2022-05-24
Update to scheduled pipeline

6l20:
## PR description

- Replace scheduled workflow with scheduled pipeline


6l20:
Done for maintenance branch only

PR #1027, opened 2022-05-26
Message format response 

darioAnongba:
Message format response.

- Remove `ethereum` types (protobuf) to use entities (Receipt and Log entities)
- Create infra type used to retrieve the data in the ethclient and unmarshal it as JSON. Infra type and entity types are not the same
- Create response types (ReceiptResponse and LogResponse), very similar to the infra types but this time used to return the response to users in the service layer
- Renamed job logs to `JobLog`


src/entities/receipt.go
```diff
@@ -0,0 +1,47 @@
...
+
+	// EEA fields
+	Output         string
+	PrivateFrom    string
+	PrivateFor     []string
+	PrivacyGroupID string
+	CommitmentHash common.Hash
+}
+
+type Log struct {
```
ggarri, line 35: I would rather rename this for something more descriptive, such as `EventLog` ReceiptLog` or `EthereumLog`

src/api/service/types/job.go
```diff
@@ -50,21 +48,22 @@ func (req *JobUpdateMessageRequest) ToJobEntity() *entities.Job {
 }
 
 type JobResponse struct {
-	UUID          string                 `json:"uuid" example:"b4374e6f-b28a-4bad-b4fe-bda36eaf849c"`                    // UUID of the job.
```
ggarri, line 53: As remark, those comments after `//` are taking automatically by swag to generate the docs for this specific object properties. Removing them from jobs it is fine because we won't be supporting them anymore in future version, but we should keep in mind we must start adding this `// {property description` next to every Request/Response object we have in `/service/types`

src/api/service/types/tx_transfer.go
```diff
@@ -8,21 +8,21 @@ import (
...
 type TransferRequest struct {
-	ChainName string            `json:"chain" validate:"required" example:"myChain"` // Name of the chain on which to send the transaction.
-	Labels    map[string]string `json:"labels,omitempty"`                            // List of custom labels.
+	ChainName string            `json:"chain" validate:"required" example:"myChain"`
+	Labels    map[string]string `json:"labels,omitempty"`
 	Params    TransferParams    `json:"params" validate:"required"`
 }
 
 type TransferParams struct {
-	Value           *hexutil.Big      `json:"value" validate:"required" example:"0x59682f00" swaggertype:"string"`                                               // Value transferred, in Wei.
```
ggarri, line 17: Revert the comments added by Pliny

src/api/service/types/tx_transfer.go
```diff
@@ -8,21 +8,21 @@ import (
 )
 
 type TransferRequest struct {
-	ChainName string            `json:"chain" validate:"required" example:"myChain"` // Name of the chain on which to send the transaction.
```
ggarri, line 11: Revert the comments added by Pliny

src/infra/ethclient/ethclient.go
```diff
@@ -18,20 +17,18 @@ import (
 
 // TransactionSender is a service for sending transaction to a blockchain
 type TransactionSender interface {
-	// SendTransaction injects a signed transaction into the pending pool for execution.
-	SendTransaction(ctx context.Context, url string, args *types.SendTxArgs) (ethcommon.Hash, error)
```
ggarri, line 22: how is it possible we were not using this method?

src/infra/ethclient/rpc/eea.go
```diff
@@ -109,41 +85,22 @@ func (ec *Client) PrivateTransactionReceipt(ctx context.Context, endpoint string
...
 	// In case private receipt is not available, we return the public receipt
 	if err != nil && errors.IsInvalidParameterError(err) {
-		return r, err
+		return r.ToEntity(), err
 	} else if err != nil {
 		return nil, errors.FromError(err).ExtendComponent(component)
 	}
 
-	// Once we have both receipts, we create a hybrid version as follow
-	r.Status, _ = hexutil.DecodeUint64(pr.Status)
```
ggarri, line 123: You missed this parsing which only applies in case of EEA receipts

ggarri (approved):

src/api/business/use-cases/contracts/decode_event_log.go
```diff
@@ -36,34 +33,21 @@ func NewDecodeEventLogUseCase(db store.DB, getContractEventsUC usecases.GetContr
...
 
-func (uc *decodeEventLogUseCase) Execute(ctx context.Context, chainUUID string, eventLog *ethereum.Log) (*ethereum.Log, error) {
+func (uc *decodeEventLogUseCase) Execute(ctx context.Context, chainUUID string, eventLog *entities.Log) (*entities.Log, error) {
 	chain, der := uc.db.Chain().FindOneByUUID(ctx, chainUUID, []string{multitenancy.WildcardTenant}, multitenancy.WildcardOwner)
 	if der != nil {
 		return nil, errors.FromError(der).ExtendComponent(decodeLogsComponent)
 	}
 
-	// This scenario is not supposed to happen
-	if len(eventLog.GetTopics()) == 0 {
```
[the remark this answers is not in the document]
darioAnongba: Impossible as per the Ethereum yellow paper. Not our responsibility if a network breaks the standard.

src/infra/ethclient/ethclient.go
darioAnongba: we are sending raw transactions, which is the purpose of orchestrate since we either sign or receive an already signed tx.

src/entities/receipt.go
darioAnongba: done

src/infra/ethclient/rpc/eea.go
darioAnongba: that's not what the docs say :thinking: https://besu.hyperledger.org/en/stable/Reference/API-Objects/#transaction-receipt-object

src/api/service/types/job.go
darioAnongba: ok, adding back

PR #1034, opened 2022-05-27
Remove formatters

darioAnongba:
Remove formatters and changes tx-listener consumer to separate message types and pass the chain directly to create subscriptions


src/api/business/use-cases/event_streams/suspend_subscriptions.go
```diff
@@ -41,7 +43,10 @@ func (uc *suspendSubscriptionsUseCase) Execute(ctx context.Context, eventStreamU
 	}
 
 	for _, sub := range subscriptions {
-		err = uc.messenger.DeleteSubscriptionMessage(ctx, sub, userInfo)
+		err = uc.messenger.DeleteSubscriptionMessage(ctx, &types.DeleteSubscriptionMessageRequest{
```
ggarri, line 46: tx-listener must be able to differentiate if a subscription was removed or suspended in order to keep track of the `fromBlock`

ggarri (approved):

src/tx-listener/service/listener/subscription.go
```diff
@@ -0,0 +1,83 @@
+package listener
```
ggarri, line 1: file was recreated

src/tx-listener/service/listener/job.go
```diff
@@ -0,0 +1,89 @@
+package listener
```
ggarri, line 1: file was recreated instead of moved

PR #1038, opened 2022-05-31
Error handling on dependency failures

darioAnongba:
Error handling on dependency failures


cmd/api/account.go
```diff
@@ -65,24 +66,24 @@ func newAccountCmd() *cobra.Command {
...
 	}
 
 	var queryInsertItems []string
 	for _, accountID := range accounts {
 		acc, err2 := client.GetEthAccount(ctx, storeName, accountID)
 		if err2 != nil {
-			log.WithField("account_id", accountID).WithError(err2).Error("Could not get account")
-			return err2
+			log.WithField("account_id", accountID).WithError(err2).Error("could not get account")
+			return err
```
ggarri, line 83: it should return err2

pkg/sdk/client/response.go
```diff
@@ -53,8 +46,7 @@ func parseResponse(ctx context.Context, response *http.Response, resp interface{
 	// Read body
 	respMsg, err := ioutil.ReadAll(response.Body)
 	if err != nil {
-		log.FromContext(ctx).WithError(err).Error(cannotReadResponseBody)
-		return errors.ServiceConnectionError(cannotReadResponseBody)
+		return err
```
ggarri, line 49: we should rephrase error here, otherwise we will respond users directly about internal GOlang error messages

pkg/sdk/client/response.go
```diff
@@ -36,15 +30,14 @@ func (m *HTTPErr) Code() int {
...
+func parseResponse(response *http.Response, resp interface{}) error {
 	if response.StatusCode == http.StatusAccepted || response.StatusCode == http.StatusOK {
 		if resp == nil {
 			return nil
 		}
 
 		if err := json.NewDecoder(response.Body).Decode(resp); err != nil {
-			log.FromContext(ctx).WithError(err).Error(invalidResponseBody)
-			return errors.ServiceConnectionError(invalidResponseBody)
+			return err
```
ggarri, line 40: we should rephrase error here, otherwise we will respond users directly about internal GOlang error messages

src/api/business/use-cases/event_streams/create.go
```diff
@@ -12,8 +11,6 @@ import (
 	"github.com/consensys/orchestrate/src/entities"
 )
 
-const createEventStreamComponent = "use-cases.create-event_stream"
```
ggarri, line 15: I still see value in having constants with the ID of the component at the top for readabilty

src/api/business/use-cases/jobs/resend_job_tx.go
```diff
@@ -35,13 +32,13 @@ func (uc *resendJobTxUseCase) Execute(ctx context.Context, jobUUID string, userI
...
 	if err != nil {
-		return errors.FromError(err).ExtendComponent(resendJobTxComponent)
+		return err
 	}
 
 	job.InternalData.ParentJobUUID = jobUUID
 	err = uc.txSenderMessenger.StartedJobMessage(ctx, job, userInfo)
 	if err != nil {
-		logger.WithError(err).Error("failed to send resend job envelope")
+		uc.logger.WithError(err).Error("failed to send resend job envelope")
```
ggarri, line 41: you should update the lines above 29-31 to create a logger with the unique fields to identify the actions
```
	ctx = log.WithFields(ctx, log.Field("job", jobUUID))
	logger := uc.logger.WithContext(ctx)
	logger.Debug("resending job transaction")
	```

File: src/api/service/types/job.go [hunk 6]
```diff
@@ -70,11 +70,11 @@ func (r *UpdateJobRequest) ToEntity(uuid string) *entities.Job {
...
 
 type JobUpdateMessageRequest struct {
-	JobUUID      string                   `json:"jobUUID,omitempty"`
-	InternalData *entities.InternalData   `json:"internal_data,omitempty"`
+	JobUUID      string                   `json:"jobUUID" validate:"required" example:"b4374e6f-b28a-4bad-b4fe-bda36eaf849c"`
+	InternalData *entities.InternalData   `json:"internalData,omitempty"`
 	Transaction  *entities.ETHTransaction `json:"transaction,omitempty"`
 	Receipt      *entities.Receipt        `json:"receipt,omitempty"`
-	Status       entities.JobStatus       `json:"status,omitempty" validate:"isJobStatus" example:"MINED"`
+	Status       entities.JobStatus       `json:"status" validate:"required,isJobStatus" example:"MINED"`
```
ggarri, line 77: we do not always update the status on the request, for instance tx-sentry updates only the internalData to attach`hasBeenRetried`

src/infra/ethclient/rpc/client.go
```diff
@@ -22,104 +20,39 @@ import (
...
-		},
-	)
+	return ec.call(req, processResult)
 }
 
 func (ec *Client) call(req *http.Request, processResult ParseResultFunc) error {
 	resp, err := ec.do(req)
 	if err != nil {
-		return err
+		return errors.EthereumClientError(err.Error())
```
ggarri, line 47: if we do this, the proxy not found 404 will lead to infinitive loop when a chain is removed. We should parse errors right below in the `ec.do()`

src/infra/ethclient/rpc/client.go
```diff
@@ -128,15 +61,35 @@ func (ec *Client) call(req *http.Request, processResult ParseResultFunc) error {
...
+		}
+
+		return errors.EthereumClientError(respMsg.Error.Message)
 	case err == nil && len(respMsg.Result) == 0:
 		return errors.NotFoundError("data not found")
 	case resp.StatusCode == 404:
-		return errors.HTTPConnectionNotFoundError("url %s not found", req.URL)
+		return errors.HTTPConnectionNotFoundError("url not found: %s", req.URL)
+	case resp.StatusCode == 429:
+		return errors.EthereumClientError("rate limiter reached")
```
ggarri, line 86: this still should be a TooManyRequest error, so we decide to retry it on the service listener level.

pkg/toolkit/app/http/client.go
```diff
@@ -34,10 +34,6 @@ func NewClient(cfg *Config) *http.Client {
 		middlewares = append(middlewares, transport.NewXAPIKeyHeadersTransport(cfg.XAPIKey))
 	}
 
-	if cfg.RetryAfterEnabled {
```
ggarri, line 37: I believe we cannot remove this as middleware of our HTTP client. The reason is we use a ratelimitter which requires to read response headers `retry-after` where information regarding when we should perform the next retry are included  https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429

src/infra/ethclient/rpc/client.go
ggarri: In addition, the comment about 429 might need to be retried on the HTTP client

src/infra/kafka/sarama/consumer_group.go
```diff
@@ -27,7 +27,7 @@ func NewConsumerGroup(cfg *Config) (*ConsumerGroup, error) {
 
 	g, err := sarama.NewConsumerGroup(cfg.URLs, cfg.GroupName, saramaCfg)
 	if err != nil {
-		return nil, errors.KafkaConnectionError(err.Error())
+		return nil, err
```
ggarri, line 30: should we parse this as InternalError?

src/infra/kafka/sarama/producer.go
```diff
@@ -43,7 +43,7 @@ func (p *Client) Send(body interface{}, topic, partitionKey string, headers map[
 
 	bValue, err := json.Marshal(body)
 	if err != nil {
-		return errors.EncodingError("failed to marshall message body")
+		return err
```
ggarri, line 46: We should only throw internal type errors from infra, this will expose an underline raw error to above layers

src/infra/postgres/gopg/parsers.go
```diff
@@ -14,15 +14,5 @@ func parseErrorResponse(err error) error {
...
 	}
 
-	pgErr, ok := err.(pg.Error)
-	if !ok {
-		return errors.PostgresConnectionError(err.Error())
-	}
-
-	switch {
-	case pgErr.IntegrityViolation():
-		return errors.ConstraintViolatedError(pgErr.Error())
```
ggarri, line 24: We should not retry on IntegratyViolation, this won't ever be solved by retrying

src/notifier/service/transaction.go
```diff
@@ -4,35 +4,31 @@ import (
...
-		logger:     log.NewLogger().SetComponent(messageListenerComponent),
 	}
 }
 
 func (mch *TransactionHandler) HandleTransactionReq(ctx context.Context, msg *entities.Message) error {
 	req := &types.TransactionMessageRequest{}
 	err := api.UnmarshalBody(bytes.NewReader(msg.Body), req)
 	if err != nil {
-		return errors.InvalidFormatError("invalid transaction request type")
+		return err
```
ggarri, line 31: we should always parse raw errors into internal type errors

src/tx-listener/integration-tests/environment.go
```diff
@@ -87,7 +85,7 @@ func NewIntegrationEnvironment(ctx context.Context, cancel context.CancelFunc, t
 		"--kafka-url=" + kafkaExternalHostname,
 		"--api-url=" + apiURL,
 		"--kafka-num-consumers=1",
-		"--log-level=panic",
+		"--log-level=debug",
```
ggarri, line 88: revert before merge

src/tx-sender/service/jobs.go
```diff
@@ -50,22 +50,14 @@ func (mch *JobHandler) HandleStartedJob(ctx context.Context, msg *entities.Messa
 			// Exits if not errors
 			case err == nil:
 				return nil
-			// if error is caused by terminated context, exit
-			case err == context.DeadlineExceeded || err == context.Canceled:
```
ggarri, line 54: Revert this, otherwise you wil be updating to FAILED jobs on context cancelled

src/tx-sender/tx-sender/nonce/manager/manager.go
```diff
@@ -73,19 +71,15 @@ func (nc *Manager) CleanNonce(ctx context.Context, job *entities.Job, jobErr err
 		return nil
 	}
 
-	// TODO: update EthClient to process and standardize nonce too low errors
-	if !strings.Contains(strings.ToLower(jobErr.Error()), "nonce too low") &&
```
ggarri, line 77: Are you doing same parsing now at the ethClient?

cmd/api/account.go
```diff
@@ -5,21 +5,22 @@ import (
...
 	"strings"
 
+	quorumkeymanager "github.com/consensys/orchestrate/src/infra/quorum-key-manager"
+
 	"github.com/consensys/orchestrate/src/infra/postgres"
 	"github.com/consensys/orchestrate/src/infra/postgres/gopg"
 
 	"github.com/consensys/orchestrate/cmd/flags"
-	"github.com/consensys/orchestrate/src/infra/quorum-key-manager/http"
+	"github.com/consensys/orchestrate/src/infra/quorum-key-manager/consensys"
```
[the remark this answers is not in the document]
darioAnongba: because I actually implemented it, before we were using the interface from QKM and we were directly using the SDK. Now I wrapped it in infra like other dependencies and gave it a name that is similar to other names in infra:
- `redigo`, `sarama`, `ristretto`, etc. `http` is not a name that follows the same philosophy for infra.

src/api/business/use-cases/event_streams/create.go
darioAnongba: This is actually good not to define a constant because it's used in only one place and it avoids reusing it in other UCs by mistake on copy-paste and actually storing that string in memory. I found many wrong components used in different places.

src/infra/ethclient/rpc/client.go
darioAnongba: I have a TooManyRequestsError defined but it's currently unused. This is supposed to be temporary until we decide on retry strategy of Ethclients. I can directly remove the `case` completely or add a TODO

src/infra/postgres/gopg/parsers.go
darioAnongba: yes we should, ConstraintViolation is a DependencyError because we do not support the current schema. Meaning there is a incoherence between the schema and the way we create/access resources. Can be a migration issue or something similar.

src/notifier/service/transaction.go
darioAnongba: as you can see, the err is lost here. We should return the actual error. This is correct in HTTP and in Kafka as well because it's a 4xx error.

src/tx-sender/tx-sender/nonce/manager/manager.go
darioAnongba: yes

pkg/toolkit/app/http/client.go
darioAnongba: but we do not use HTTP for intra-communication betweens MSs now. We use Kafka and the tx-sentry is the only exception and we know that we need to refactor it eventually to remove dependence on the API. The Go HTTP client will only be used in tests. I see it as an SDK like the JS SDK.

cmd/api/account.go
darioAnongba: done

pkg/sdk/client/response.go
darioAnongba: no, this is correct. the SDK is on the user side. Only the tx-sentry is using it now and that should be removed. The SDK is now a "real" SDK that is not used by orchestrate. It was actually a big mistake to use `errors` pkg in the SDK.

pkg/sdk/client/response.go
darioAnongba: same response

pkg/sdk/client/response.go
```diff
@@ -98,36 +90,33 @@ func parseResponseError(statusCode int, errMsg string) error {
...
 		}
 
 		return "", NewHTTPErr(errors.Errorf(errResp.Code, errResp.Message), response.StatusCode)
 	}
 
 	responseData, err := ioutil.ReadAll(response.Body)
 	if err != nil {
-		log.FromContext(ctx).WithError(err).Error(invalidResponseBody)
-		return "", errors.ServiceConnectionError(invalidResponseBody)
+		return "", err
```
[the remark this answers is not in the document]
darioAnongba: same

src/api/business/use-cases/jobs/resend_job_tx.go
darioAnongba: done

src/api/service/types/job.go
darioAnongba: done

src/infra/kafka/sarama/consumer_group.go
darioAnongba: no it's fine because this is the initialization of the infra component, in the `New` functions we don't need to return any special type. We used to return `ConfigError` but that wasn't needed since we never check them. this will make the app fail to start so we want the actual error.

src/infra/kafka/sarama/producer.go
darioAnongba: done

src/tx-listener/integration-tests/environment.go
darioAnongba: done

src/infra/ethclient/rpc/client.go
darioAnongba: reverted

ggarri (approved):

PR #1045, opened 2022-06-03
Fix e2e + upgrade packages

darioAnongba:
## Documentation

## Changelog


tests/e2e/account_test.go
```diff
@@ -120,20 +121,20 @@ func (s *accountTestSuite) TestAccounts_FailureScenarios() {
...
 		require.NoError(t, err)
-		
+
 		_, err = s.env.Client.CreateAccount(s.ctx, fakeReq)
 		require.Error(t, err)
 		assert.Equal(t, http.StatusConflict, err.(*client.HTTPErr).Code())
 	})
-	
+
 	s.T().Run("when an user creates an account with a not existing QKM StoreID if should fail with expected error", func(t *testing.T) {
```
6l20, line 130: typo, "it should" instead of "if should"

6l20 (approved):

tests/e2e/account_test.go
```diff
@@ -120,20 +121,20 @@ func (s *accountTestSuite) TestAccounts_FailureScenarios() {
...
 		fakeReq := testdata.FakeCreateAccountRequest()
 		fakeReq.StoreID = "notExisting"
 		_, err := s.env.Client.CreateAccount(s.ctx, fakeReq)
 		require.Error(t, err)
-		assert.Equal(t, http.StatusFailedDependency, err.(*client.HTTPErr).Code())
+		assert.Equal(t, http.StatusNotFound, err.(*client.HTTPErr).Code())
```
ggarri, line 135: is not found the expected error code for this? As far as I see it NotFound should be returned in case you query for an specific asset and it is not found, but in case you create a new element if one of its dependencies it is not found we should expect a BadRequest

tests/e2e/account_test.go
```diff
@@ -142,6 +143,6 @@ func (s *accountTestSuite) TestAccounts_FailureScenarios() {
 		notExistingAddr := testdata2.FakeAddress()
 		_, err := s.env.Client.SignMessage(s.ctx, *notExistingAddr, signMsgReq)
 		require.Error(t, err)
-		assert.Equal(t, http.StatusBadRequest, err.(*client.HTTPErr).Code())
+		assert.Equal(t, http.StatusNotFound, err.(*client.HTTPErr).Code())
```
ggarri, line 146: I believe BadRequest it was the proper error code

PR #1064, opened 2022-06-13
Add event streams in transactions

darioAnongba:
Add event streams in transactions


src/api/business/use-cases/accounts/fund_account.go
```diff
@@ -42,13 +40,13 @@ func (uc *fundAccountUseCase) Execute(ctx context.Context, account *entities.Acc
 
 	chains, err := uc.searchChainsUC.Execute(ctx, &entities.ChainFilters{Names: []string{chainName}}, userInfo)
 	if err != nil {
-		return errors.FromError(err).ExtendComponent(fundAccountComponent)
+		return err
 	}
 
 	if len(chains) == 0 {
 		errMsg := "chain does not exist"
 		logger.Warn(errMsg)
```
ggarri, line 48: I think we can remove this Warn message, there is not value in it

src/api/business/use-cases/event_streams/create.go
```diff
@@ -44,38 +42,6 @@ func (uc *createUseCase) Execute(ctx context.Context, eventStream *entities.Even
 		return nil, errors.AlreadyExistsError(errMsg)
 	}
 
-	if chainName == entities.WildcardChainName {
```
ggarri, line 47: This reminds me you should update the migration to remove the UNIQUE index on chain

src/api/business/use-cases/event_streams/delete.go
```diff
@@ -33,7 +36,18 @@ func (uc *deleteUseCase) Execute(ctx context.Context, uuid string, userInfo *mul
 		return err
 	}
 
-	err = uc.db.Delete(ctx, uuid, userInfo.AllowedTenants, userInfo.Username)
+	subs, err := uc.db.Subscription().Search(ctx, &entities.SubscriptionFilters{EventStreamUUID: uuid}, userInfo.AllowedTenants, userInfo.Username)
+	if err != nil {
+		return err
+	}
+
+	if len(subs) > 0 {
```
ggarri, line 44: aren't we failing also on PENDING jobs?

ggarri (approved):

src/api/business/use-cases/accounts/fund_account.go
darioAnongba: removed

Issue #1037, opened 2022-05-31
SearchAccounts crashes in case of large list of available accounts


ggarri:
### Describe the bug

Orchestrate API closes HTTP connection in case querying for all the accounts in the list is larger than 1M

### Configuration

Deploy and Orchestrate environment with 1M keys

### Steps to reproduce

1. Run `/accounts/search`

**Actual result**
500 HTTP error

**Expected result**
200 

## Acceptance criteria
- [ ] Orchestrate support account pagination
- [ ] Orchestrate support transaction pagination
- [ ] SDK pagination support

PR #1066, opened 2022-06-15
Fix/search accounts

6l20:
## PR description

- Add limit and page to search accounts request
- Add limit and page to search transactions request
- Update sdk

## Fixed Issue(s)
#1037 

## Documentation

## Changelog


src/api/business/use-cases/accounts/search_accounts_test.go
```diff
@@ -33,6 +33,7 @@ func TestSearchAccounts_Execute(t *testing.T) {
 
 		filter := &entities.AccountFilters{
 			Aliases: []string{"alias1"},
+			Pagination: entities.PaginationFilters{Limit: 10, Offset: 100},
```
ggarri, line 36: It seems like this change does not have the corresponding test to validate the size of the response it is the expected

src/api/integration-tests/accounts.go
```diff
@@ -183,6 +183,40 @@ func (s *accountsTestSuite) TestSearch() {
...
+		txRequest0 := testdata.FakeCreateAccountRequest()
+
+		_, err := s.client.CreateAccount(ctx, txRequest0)
+		require.NoError(s.T(), err)
+		txRequest1 := testdata.FakeCreateAccountRequest()
+
+		ethAccRes1, err := s.client.CreateAccount(ctx, txRequest1)
+		resp, err := s.client.SearchAccounts(ctx, &entities.AccountFilters{
+			Aliases: []string{txRequest0.Alias, txRequest1.Alias},
+			Pagination: entities.PaginationFilters{Limit: 1, Offset: 1},
```
ggarri, line 197: This test unit appears to be depending on the order and actions of other test units, for instance if we add a new test just above which add another account, it would make this test to fail. Every test unit MUST be independent

src/infra/api/format.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"net/http"
+	"strconv"
+
+	"github.com/consensys/orchestrate/pkg/errors"
+	"github.com/consensys/orchestrate/src/entities"
+)
+
+func FilterIntegerValueWithKey(req *http.Request, key string, filters *entities.PaginationFilters) (*entities.PaginationFilters, error) {
```
ggarri, line 11: This function holds a lot of complexity and edges cases you didn't cover such as:
- if key dis not "limit" or "offset"
- if a value is already set
- Does not validate the consistency of the values


This method can be easily implemented like the following:
```
func PaginationFilter(req *http.Request) (*entities.PaginationFilters, error) {
filters = &entities.PaginationFilters{}
qOffset := req.URL.Query().Get("offset")
if qOffset != "" {
  filters.offset , err := strconv.Atoi(qOffset)
   if err != nil {...}
}
qLimit := req.URL.Query().Get("limit")
if qLimit != "" {
  filters.limit , err := strconv.Atoi(qLimit)
   if err != nil {...}
}
return filters
}

File: pkg/sdk/client/transactions.go [hunk 4]
```diff
@@ -131,3 +135,37 @@ func (c *HTTPClient) SendSpeedUpTransaction(ctx context.Context, txRequestUUID s
...
+func (c *HTTPClient) SearchTransaction(ctx context.Context, filters *entities.TransactionFilters) ([]*types.TransactionResponse, error) {
+	reqURL := fmt.Sprintf("%v/transactions", c.config.URL)
+	var resp []*types.TransactionResponse
+
+	var qParams []string
+	if len(filters.IdempotencyKeys) > 0 {
+		qParams = append(qParams, "idempotency_keys="+strings.Join(filters.IdempotencyKeys, ","))
+	}
+
+	if filters.Pagination.Limit > 0 {
```
ggarri, line 148: Pagination logic should be a bit more advantage to facilitate its usage. For instance, Stripe provides a good example of this https://stripe.com/docs/api/pagination/search

src/api/integration-tests/accounts.go
6l20: Here it is made independent thanks to the Aliases which are unique on the whole test set

pkg/sdk/client/transactions.go
```diff
@@ -131,3 +135,37 @@ func (c *HTTPClient) SendSpeedUpTransaction(ctx context.Context, txRequestUUID s
...
+		qParams = append(qParams, "limit="+strconv.Itoa(filters.Pagination.Limit))
+	}
+
+	if filters.Pagination.Page > 0 {
```
ggarri, line 152: page 0 is valid, it means first page

src/api/business/use-cases/accounts/search_accounts.go
```diff
@@ -31,6 +31,6 @@ func (uc *searchAccountsUseCase) Execute(ctx context.Context, filters *entities.
 		return nil, errors.FromError(err).ExtendComponent(searchAccountsComponent)
 	}
 
-	uc.logger.WithContext(ctx).Trace("accounts found successfully")
+	uc.logger.WithContext(ctx).Debug("accounts found successfully")
```
ggarri, line 34: Revert this to trace

src/api/business/use-cases/accounts/search_accounts_test.go
```diff
@@ -33,13 +33,15 @@ func TestSearchAccounts_Execute(t *testing.T) {
...
 			Aliases: []string{"alias1"},
+			Pagination: entities.PaginationFilters{Limit: 10, Page: 100},
 		}
 
 		accountAgent.EXPECT().Search(gomock.Any(), filter, userInfo.AllowedTenants, userInfo.Username).Return([]*entities.Account{acc}, nil)
 
 		resp, err := usecase.Execute(ctx, filter, userInfo)
 
 		assert.NoError(t, err)
+		assert.Len(t, resp, 1)
```
ggarri, line 44: what does this test cover?

src/api/service/controllers/accounts.go
```diff
@@ -127,23 +129,49 @@ func (c *AccountsController) search(rw http.ResponseWriter, request *http.Reques
...
 	accs, err := c.ucs.Search().Execute(ctx, filters, multitenancy.UserInfoValue(ctx))
 	if err != nil {
 		infra.WriteHTTPErrorResponse(rw, err)
 		return
 	}
 
-	response := []*api.AccountResponse{}
+	// signal when more items are available
+	hasMore := false
+	if filters.Pagination.Limit > 0 {
```
ggarri, line 157: can limit be ever less than `0`?

src/api/service/controllers/accounts.go
```diff
@@ -127,23 +129,49 @@ func (c *AccountsController) search(rw http.ResponseWriter, request *http.Reques
 	rw.Header().Set("Content-Type", "application/json")
 	ctx := request.Context()
 
-	filters, err := formatAccountFilterRequest(request)
+	// Default filter to limit of defaultLimit + 1 items
+	var filters *entities.AccountFilters
+	filters = &entities.AccountFilters{}
+	filters.Pagination = entities.PaginationFilters{Limit: api.DefaultAccountPageSize}
```
ggarri, line 135: The default value should be set in case it is not set, not the other way around. Otherwise it leads to confusion

pkg/sdk/client/transactions.go
6l20: Yes but in that case it is not passed in the query, only non zero are passed

pkg/sdk/client/accounts.go
```diff
@@ -83,15 +84,23 @@ func (c *HTTPClient) UpdateAccount(ctx context.Context, address ethcommon.Addres
...
 	var qParams []string
 	if len(filters.Aliases) > 0 {
 		qParams = append(qParams, "aliases="+strings.Join(filters.Aliases, ","))
 	}
 
+	if filters.Pagination.Limit > 0 {
+		qParams = append(qParams, "limit="+strconv.Itoa(filters.Pagination.Limit))
+	}
+
+	if filters.Pagination.Page > 0 {
```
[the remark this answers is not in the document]
6l20: No, page 0 is not needed in the query

src/api/business/use-cases/accounts/search_accounts_test.go
6l20: This is useless indeed

pkg/sdk/client/accounts.go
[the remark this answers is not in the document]
6l20: page 0 is an implied value, it should not be explicitly set, so pages in query always start at 1

src/api/business/use-cases/accounts/search_accounts.go
6l20: done

src/api/service/controllers/accounts.go
6l20: To me this is more like a declaration, filter config, and it is better done at the start, I see no confusion here

src/api/business/use-cases/accounts/search_accounts_test.go
6l20: Length related tests are all located in integration

src/api/service/controllers/accounts.go
ggarri: then why you need this?

src/api/service/controllers/accounts.go
ggarri: you then will not know if it was replaced or not in the following statements

6l20: Why would you need to know ?

src/api/service/controllers/accounts_test.go
```diff
@@ -268,6 +268,87 @@ func (s *accountsCtrlTestSuite) TestAccountController_SearchIdentity() {
...
+
+		accounts := []*entities.Account{accResp0, accResp1, accResp2, accResp3}
+		rw := httptest.NewRecorder()
+
+		httpRequest := httptest.
+			NewRequest(http.MethodGet, "/accounts?limit=4", nil).
+			WithContext(s.ctx)
+
+		filter := &entities.AccountFilters{
+			Pagination: entities.PaginationFilters{Limit: 4},
```
ggarri: since we are using `hash_more`, shouldn't limit be `5` ?

src/api/service/controllers/accounts_test.go
```diff
@@ -268,6 +268,87 @@ func (s *accountsCtrlTestSuite) TestAccountController_SearchIdentity() {
...
+		accResp0 := testdata.FakeAccount()
+		accResp1 := testdata.FakeAccount()
+		accResp2 := testdata.FakeAccount()
+		accResp3 := testdata.FakeAccount()
+
+		accounts := []*entities.Account{accResp0, accResp1, accResp2, accResp3}
+		rw := httptest.NewRecorder()
+
+		httpRequest := httptest.
+			NewRequest(http.MethodGet, "/accounts?offset=100", nil).
```
ggarri: aren't we using `page` instead of `offset` as input?

src/api/service/controllers/transactions.go
```diff
@@ -242,6 +242,8 @@ func (c *TransactionsController) getOne(rw http.ResponseWriter, request *http.Re
 // @Security     ApiKeyAuth
 // @Security     JWTAuth
 // @Param        idempotency_keys  query     []string                 false  "List of idempotency keys"  collectionFormat(csv)
+// @Param		 limit	  query     int					 false  "Maximum response size"
+// @Param		 offset	  query     int					 false  "Offset within the entire responses set"
```
ggarri: should be `page`

src/api/integration-tests/accounts.go
ggarri: How this test is independent from other? you are adding an element and then fetching it with limit 1 and offset 1, what it means you expect this element to be added in second position. If I copy paste this test and move it to the bottom of the file it won't pass, that is why I know this test does not run independently from other

src/api/service/controllers/accounts.go
```diff
@@ -118,6 +118,8 @@ func (c *AccountsController) getOne(rw http.ResponseWriter, request *http.Reques
 // @Security     ApiKeyAuth
 // @Security     JWTAuth
 // @Param        aliases  query     []string             false  "List of account aliases"  collectionFormat(csv)
+// @Param		 limit	  query     int					 false  "Maximum response size"
+// @Param		 offset	  query     int					 false  "Offset within the entire responses set"
```
ggarri: page

ggarri (approved):

PR #1070, opened 2022-06-16
889 listening depth

ggarri:
## Documentation

## Changelog


src/tx-listener/tx-listener/use-cases/subscriptions/notify_events.go
```diff
@@ -35,11 +35,23 @@ func NotifyEventsUseCase(messenger sdk.MessengerAPI, proxyClient sdk.ChainProxyC
 	}
 }
 
-func (uc *notifyEventsUC) Execute(ctx context.Context, sub *entities.Subscription) error {
+func (uc *notifyEventsUC) Execute(ctx context.Context, sub *entities.Subscription, chain *entities.Chain) error {
 	logger := uc.logger.WithField("subscription", sub.UUID)
 
+	var toBlock *big.Int
+
+	if chain.ListenerDepth > 0 {
```
darioAnongba, line 43: I don't understand this implementation, isn't the `depth` just the number of blocks to need to wait to consider a receipt or event valid? How is this implementatio working?

src/tx-listener/tx-listener/use-cases/jobs/pending.go
```diff
@@ -36,7 +37,7 @@ func PendingJob(proxyClient sdk.ChainProxyClient,
 	}
 }
 
-func (uc *pendingJobMsg) Execute(ctx context.Context, job *entities.Job) (*entities.Receipt, error) {
+func (uc *pendingJobMsg) Execute(ctx context.Context, job *entities.Job, chain *entities.Chain) (*entities.Receipt, error) {
```
darioAnongba, line 40: It seems you only need the depth and not the whole chain to be passed :thinking:

src/tx-listener/tx-listener/use-cases/subscriptions/notify_events.go
ggarri: Rewrote logic for better clarity. We discard events if the distant to latest block is lowe the chain depth

src/tx-listener/tx-listener/use-cases/jobs/pending.go
ggarri: indeed, that is an optimization I can do.

darioAnongba (approved):

PR #1073, opened 2022-06-16
added speed up and call off endpoints

ggarri:
## Documentation

## Changelog


services/api/service/controllers/transactions.go
```diff
@@ -269,3 +276,69 @@ func (c *TransactionsController) search(rw http.ResponseWriter, request *http.Re
...
+	if err != nil {
+		httputil.WriteHTTPErrorResponse(rw, err)
+		return
+	}
+
+	_ = json.NewEncoder(rw).Encode(formatters.FormatTxResponse(tx))
+}
+
+// @Summary      Call of transaction
+// @Description  Call of transaction by sending empty data transaction with same nonce and additional 10% more gas
```
6l20, line 322: Should be "Call off" instead of "Call of"

6l20 (approved):

services/api/service/controllers/transactions.go
ggarri: done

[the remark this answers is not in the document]
ggarri: done

PR #1089, opened 2022-06-28
961 ext tx listener

ggarri:
## PR description
- Support listening of external transactions

## Documentation

## Changelog


pkg/sdk/client.go
```diff
@@ -41,6 +41,7 @@ type TransactionClient interface {
 	SendDeployTransaction(ctx context.Context, request *types.DeployContractRequest) (*types.TransactionResponse, error)
 	SendRawTransaction(ctx context.Context, request *types.RawTransactionRequest) (*types.TransactionResponse, error)
 	SendTransferTransaction(ctx context.Context, request *types.TransferRequest) (*types.TransactionResponse, error)
+	TxHashTransaction(ctx context.Context, request *types.TxHashRequest) (*types.TransactionResponse, error)
```
6l20, line 44: Method name could follow the verb prefix pattern used elsewhere

src/api/business/use-cases/jobs/start_job_test.go
```diff
@@ -25,8 +26,8 @@ func TestStartJob_Execute(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
 
-	jobDA := mocks.NewMockJobAgent(ctrl)
-	messenger := mock3.NewMockMessengerTxSender(ctrl)
+	txSenderMessenger := mock3.NewMockMessengerTxSender(ctrl)
```
6l20, line 29: We could have a name that is more explicit than mock3 here

src/tx-sender/tx-sender/use-cases/sender/send-eea-private-tx.go
```diff
@@ -59,18 +56,18 @@ func (uc *sendEEAPrivateTxUseCase) Execute(ctx context.Context, job *entities.Jo
...
 
 	job.Transaction.Raw, _, err = uc.signTx.Execute(ctx, job)
 	if err != nil {
-		return errors.FromError(err).ExtendComponent(sendEEAPrivateTxComponent)
+		return err
 	}
 
 	job.Transaction.Hash, err = uc.sendTx(ctx, job)
 	if err != nil {
 		if err2 := uc.nonceManager.CleanNonce(ctx, job, err); err2 != nil {
```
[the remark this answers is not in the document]
ggarri: Good catch, fixed

src/api/business/use-cases/jobs/start_job_test.go
ggarri: This is a common practice in our unit tests which comes from the autoimport feature of our IDE. Since I see not value on this I see not reason to make efforts in renaming them as it has a single usage in the whole file

6l20: Ok with that

6l20 (approved):

